PR #6, opened 2018-08-17
adding code import references in comments

tfoote:
Referencing where the code was copied from for better traceability.


tfoote:
@gaoethan When I restructured your test here: https://github.com/gaoethan/message_filters-1/blob/32898573597eb115f3d4d6cb24aeace6e4ad3eca/test/test_fuzz.cpp I assumed that it had been copied from somewhere due to the 2008 Willow Garage copyright. However looking at it I cannot find where it came from as imported here: https://github.com/ros2/message_filters/commit/193cf7bb7860390ac84d8d2a4bb932be96d9be12#diff-3969d386a591d68079e4e7903ef87e13

Did it start as another file and you modified it?

gaoethan:
@tfoote Actually, it's more like the tests within [time_sequencer_unittest.cpp](https://github.com/ros2/message_filters/blob/master/test/time_sequencer_unittest.cpp) but with self-defined messages, you can see they have similiar code structure, thanks !

Issue #2, opened 2018-07-09
Integrate changes from intel/ros2_message_filters repository


dirk-thomas:
@jwang11 @gaoethan @testkit In order to integrate the various changes from the https://github.com/intel/ros2_message_filters repository here please create separate PRs for separate /unrelated changes:

* remove no more relevant (ROS 1 specific) code
* add additional files from other repositories (e.g. `MessageEvent`) - depending on the amount of code the history might be good to keep (?)
* refactor code from `Boost` to `std`
* update source code, tests, CMake, manifest to ament / ROS 2
* (anything I missed while looking through the "diff" in the `intel` repo)

By making separate PRs with orthogonal changes it will be easier / quicker to review / iterate /merge them.

gaoethan:
okay, let me try to split them into small patches to ease this process. thanks !

testkit:
@dirk-thomas, will you pls. grant @gaoethan write access to merge the PR?

dirk-thomas:
> will you pls. grant @gaoethan write access to merge the PR?

We are happy to grant write access to you. For now I would like to focus on iterating on the PR. As commented there it needs some updates to follow the guideline we provided in this ticket.

[The idea of granting write access is not to "just merge the PR as-is". A PR should always be reviewer from another person before it is being merged.]

gaoethan:
Now because @tfoote is working on his refactoring with the branch https://github.com/ros2/message_filters/compare/refactor4 to move the message_filters here, in order to keep his coherence and make the patches conflict potentially each other, let's go ahead after @tfoote initial refactoring on a whole as an unified base for further development/issue fix. thanks !

mjcarroll:
Is this closed by ros2/message_filters#5?

mikaelarguedas:
It looks like the python tests have not been ported over yet https://github.com/ros2/message_filters/pull/5#discussion_r206355342

sloretz:
@gaoethan It looks like https://github.com/intel/ros2_message_filters/ already has the python `message_filters` code ported to ros2. May I ask you to open a PR with the python code? If I understand correctly that being merged marks when this ticked is complete.

gaoethan:
ok. no problem, let me file a PR for the python part.

gaoethan:
I have already integrated the ros2 `message_filters `python support with some previous changes of [intel/ros2_message_filters](https://github.com/intel/ros2_message_filters) and new changes to fully support of headerless messages. please find the details underneath #7 , thanks !

[...]

PR #7, opened 2018-08-19
enable message_filters support of python interfaces and tests

gaoethan:
I ported the message_filters ROS2 python support here and it mainly involves the following changes: 
- replace Subsciber with ROS2 `Subscriber`
- use ROS2 ROSTIME to support headerless messages for `Cache` and `ApproximateTimeSynchronizer`
- use ROS2 `rclpy` root logger for logging
- enable exact timestamp policy for with `TimeSynchronizer` 
- enable approximate timestamp policy for messages with header 
- enable approximate timestamp policy test for headerless messages
- enable message cache for messages with header 
- enable message cache for for headerless messages
- fix doc and coding style issue
- enable the relevant tests with `ament_cmake_pytest`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,6 +16,7 @@ find_package(ament_cmake REQUIRED)
 find_package(rclcpp REQUIRED)
 find_package(rcutils)
 find_package(builtin_interfaces REQUIRED)
+find_package(std_msgs REQUIRED)
 find_package(sensor_msgs REQUIRED)
 
 include_directories(include)
@@ -23,6 +24,7 @@ add_library(${PROJECT_NAME} SHARED src/connection.cpp)
 ament_target_dependencies(${PROJECT_NAME}
   "rclcpp"
   "rcutils"
+  "std_msgs"
   "builtin_interfaces"
   "sensor_msgs")
 
@@ -108,6 +110,11 @@ if(BUILD_TESTING)
     target_link_libraries(${PROJECT_NAME}-test_fuzz ${PROJECT_NAME})
   endif()
 
+  # python tests with python interfaces of message filters
+  find_package(ament_cmake_pytest REQUIRED)
+  ament_add_pytest_test(directed.py "test/directed.py")
+  ament_add_pytest_test(test_approxsync.py "test/test_approxsync.py")
+  ament_add_pytest_test(test_message_filters_cache.py "test/test_message_filters_cache.py")
 endif()
 
 ament_package()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,7 +16,6 @@ find_package(ament_cmake REQUIRED)
 find_package(rclcpp REQUIRED)
 find_package(rcutils)
 find_package(builtin_interfaces REQUIRED)
-find_package(std_msgs REQUIRED)
 find_package(sensor_msgs REQUIRED)
 
 include_directories(include)
@@ -24,7 +23,6 @@ add_library(${PROJECT_NAME} SHARED src/connection.cpp)
 ament_target_dependencies(${PROJECT_NAME}
   "rclcpp"
   "rcutils"
-  "std_msgs"
   "builtin_interfaces"
   "sensor_msgs")
 
--- a/package.xml
+++ b/package.xml
@@ -19,6 +19,7 @@
   <build_depend>builtin_interfaces</build_depend>
   <build_depend>rclcpp</build_depend>
   <build_depend>rclpy</build_depend>
+  <build_depend>std_msgs</build_depend>
   <build_depend>sensor_msgs</build_depend>
 
   <test_depend>ament_lint_auto</test_depend>
--- a/package.xml
+++ b/package.xml
@@ -19,7 +19,6 @@
   <build_depend>builtin_interfaces</build_depend>
   <build_depend>rclcpp</build_depend>
   <build_depend>rclpy</build_depend>
-  <build_depend>std_msgs</build_depend>
   <build_depend>sensor_msgs</build_depend>
 
   <test_depend>ament_lint_auto</test_depend>
--- a/src/message_filters/__init__.py
+++ b/src/message_filters/__init__.py
@@ -35,6 +35,11 @@ import itertools
 import threading
 import rclpy
 
+from rclpy.clock import ROSClock
+from rclpy.duration import Duration
+from rclpy.time import Time
+
+
 class SimpleFilter(object):
 
     def __init__(self):
@@ -118,7 +123,8 @@ class Cache(SimpleFilter):
                               "Use the 'allow_headerless' constructor option to "
                               "auto-assign ROS time to headerless messages.")
                 return
-            stamp = rospy.Time.now()
+
+            stamp = ROSClock().now()
         else:
             stamp = msg.header.stamp
 
@@ -238,7 +244,7 @@ class ApproximateTimeSynchronizer(TimeSynchronizer):
 
     def __init__(self, fs, queue_size, slop, allow_headerless=False):
         TimeSynchronizer.__init__(self, fs, queue_size)
-        self.slop = rospy.Duration.from_sec(slop)
+        self.slop = Duration(seconds=slop).nanoseconds
         self.allow_headerless = allow_headerless
 
     def add(self, msg, my_queue, my_queue_index=None):
@@ -248,7 +254,8 @@ class ApproximateTimeSynchronizer(TimeSynchronizer):
                               "Use the 'allow_headerless' constructor option to "
                               "auto-assign ROS time to headerless messages.")
                 return
-            stamp = rospy.Time.now()
+
+            stamp = ROSClock().now().nanoseconds
         else:
             stamp = msg.header.stamp
 
--- a/src/message_filters/__init__.py
+++ b/src/message_filters/__init__.py
@@ -49,7 +49,8 @@ class SimpleFilter(object):
     def registerCallback(self, cb, *args):
         """
         Register a callback function `cb` to be called when this filter
-        has output. The filter calls the function ``cb`` with a filter-dependent
+        has output.
+        The filter calls the function ``cb`` with a filter-dependent
         list of arguments,followed by the call-supplied arguments ``args``.
         """
 
--- a/src/message_filters/__init__.py
+++ b/src/message_filters/__init__.py
@@ -291,7 +291,7 @@ class ApproximateTimeSynchronizer(TimeSynchronizer):
                 return
             topic_stamps = sorted(topic_stamps, key=lambda x: x[1])
             stamps.append(topic_stamps)
-        for vv in itertools.product(*[zip(*s)[0] for s in stamps]):
+        for vv in itertools.product(*[list(zip(*s))[0] for s in stamps]):
             vv = list(vv)
             # insert the new message
             if my_queue_index is not None:
--- a/src/message_filters/__init__.py
+++ b/src/message_filters/__init__.py
@@ -49,9 +49,8 @@ class SimpleFilter(object):
     def registerCallback(self, cb, *args):
         """
         Register a callback function `cb` to be called when this filter
-        has output.
-        The filter calls the function ``cb`` with a filter-dependent list of arguments,
-        followed by the call-supplied arguments ``args``.
+        has output. The filter calls the function ``cb`` with a filter-dependent
+        list of arguments,followed by the call-supplied arguments ``args``.
         """
 
         conn = len(self.callbacks)
@@ -65,10 +64,10 @@ class SimpleFilter(object):
 class Subscriber(SimpleFilter):
     
     """
-    ROS subscription filter.  Identical arguments as :class:`rospy.Subscriber`.
+    ROS2 subscription filter,Identical arguments as :class:`rclpy.Subscriber`.
 
     This class acts as a highest-level filter, simply passing messages
-    from a ROS subscription through to the filters which have connected
+    from a ROS2 subscription through to the filters which have connected
     to it.
     """
     def __init__(self, *args, **kwargs):
@@ -120,10 +119,10 @@ class Cache(SimpleFilter):
     def add(self, msg):
         if not hasattr(msg, 'header') or not hasattr(msg.header, 'stamp'):
             if not self.allow_headerless:
-                rclpy.logging._root_logger.log("can not use message filters "
-                              "for messages without timestamp infomation when "
-                              "'allow_headerless' is disabled. auto assign "
-                              "ROSTIME to headerless messages once enabling "
+                rclpy.logging._root_logger.log("can not use message filters"
+                              "for messages without timestamp infomation when"
+                              "'allow_headerless' is disabled. auto assign"
+                              "ROSTIME to headerless messages once enabling"
                               "constructor option of 'allow_headerless'.",
                               LoggingSeverity.INFO)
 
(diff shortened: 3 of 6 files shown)


gaoethan:
@tfoote @mjcarroll @dirk-thomas  maybe you have some extra feedback, please help review and kindly share, thanks !

CMakeLists.txt
```diff
@@ -16,13 +16,15 @@ find_package(ament_cmake REQUIRED)
 find_package(rclcpp REQUIRED)
 find_package(rcutils)
 find_package(builtin_interfaces REQUIRED)
+find_package(std_msgs REQUIRED)
```
tfoote: I only see this being used in a test https://github.com/ros2/message_filters/pull/7/files#diff-d26d660e83afc31871c6cfeaecd487e8R40 So this should be in a testing only block.

CMakeLists.txt
```diff
@@ -16,13 +16,15 @@ find_package(ament_cmake REQUIRED)
...
 find_package(sensor_msgs REQUIRED)
 
 include_directories(include)
 add_library(${PROJECT_NAME} SHARED src/connection.cpp)
 ament_target_dependencies(${PROJECT_NAME}
   "rclcpp"
   "rcutils"
+  "std_msgs"
```
tfoote: Like above this should only be triggered for tests. It looks like it's already in the testing block below.

package.xml
```diff
@@ -19,6 +19,7 @@
   <build_depend>builtin_interfaces</build_depend>
   <build_depend>rclcpp</build_depend>
   <build_depend>rclpy</build_depend>
+  <build_depend>std_msgs</build_depend>
```
tfoote: Like above this appears to already be a test_depend why add it here?

src/message_filters/__init__.py
```diff
@@ -43,9 +49,8 @@ def __init__(self):
     def registerCallback(self, cb, *args):
         """
         Register a callback function `cb` to be called when this filter
-        has output.
-        The filter calls the function ``cb`` with a filter-dependent list of arguments,
-        followed by the call-supplied arguments ``args``.
+        has output. The filter calls the function ``cb`` with a filter-dependent
```
tfoote: This looks like an unnecessary change of indention/whitespace only. We usually start new sentances on new lines in documentation to avoid reflowing whole paragraphs when making changes to wordings.

src/message_filters/__init__.py
```diff
@@ -113,11 +119,16 @@ def connectInput(self, f):
     def add(self, msg):
         if not hasattr(msg, 'header') or not hasattr(msg.header, 'stamp'):
             if not self.allow_headerless:
-                rospy.logwarn("Cannot use message filters with non-stamped messages. "
-                              "Use the 'allow_headerless' constructor option to "
-                              "auto-assign ROS time to headerless messages.")
+                rclpy.logging._root_logger.log("can not use message filters"
```
tfoote: Please use the node's public API to get the logger not the modules implementation. 

Here's an example: https://github.com/ros2/demos/blob/e5ea2e7fec944d08d04816ef0b8e2f89e5171880/demo_nodes_py/demo_nodes_py/topics/talker.py#L34

Also this has a lot of reflow that would be good to avoid.

src/message_filters/__init__.py
```diff
@@ -227,27 +238,32 @@ class ApproximateTimeSynchronizer(TimeSynchronizer):
...
+    that defines the delay (in seconds) with which messages can be synchronized.
+    The ``allow_headerless`` option specifies whether to allow storing
+    headerless messages with current ROS time instead of timestamp. You should
     avoid this as much as you can, since the delays are unpredictable.
     """
 
     def __init__(self, fs, queue_size, slop, allow_headerless=False):
         TimeSynchronizer.__init__(self, fs, queue_size)
-        self.slop = rospy.Duration.from_sec(slop)
+        self.slop = Duration(seconds=slop).nanoseconds
```
tfoote: Please use the semantic datatypes in the storage don't just convert it to the int values.

src/message_filters/__init__.py
```diff
@@ -227,27 +238,32 @@ class ApproximateTimeSynchronizer(TimeSynchronizer):
...
+                rclpy.logging._root_logger.log("can not use message filters"
+                              "for messages without timestamp infomation when"
+                              "'allow_headerless' is disabled. auto assign"
+                              "ROSTIME to headerless messages once enabling"
+                              "constructor option of 'allow_headerless'.",
+                              LoggingSeverity.INFO)
                 return
-            stamp = rospy.Time.now()
+
+            stamp = ROSClock().now().nanoseconds
```
tfoote: Same here, use the semantic datatypes.

src/message_filters/__init__.py
```diff
@@ -275,7 +291,7 @@ def add(self, msg, my_queue, my_queue_index=None):
                 return
             topic_stamps = sorted(topic_stamps, key=lambda x: x[1])
             stamps.append(topic_stamps)
-        for vv in itertools.product(*[zip(*s)[0] for s in stamps]):
+        for vv in itertools.product(*[list(zip(*s))[0] for s in stamps]):
```
tfoote: Why this change? The output of zip is an iterable and converting it to a list doesn't seem to add functionality.

test/test_message_filters_cache.py
```diff
@@ -55,108 +57,113 @@ def __init__(self):
...
 
         msg = AnonymMsg()
-        msg.header.stamp = rospy.Time(4)
+        msg.header.stamp = Time(seconds=4)
         cache.add(msg)
 
-        l = len(cache.getInterval(rospy.Time(2.5),
-                                  rospy.Time(3.5)))
-        self.assertEquals(l, 1, "invalid number of messages" +
+        l = len(cache.getInterval(Time(seconds=2.5), Time(seconds=3.5)))
```
tfoote: Can you keep the wrapping for a smaller diff and easier reading in this and the next 3 sections.

src/message_filters/__init__.py
gaoethan: @tfoote Actually, I've ever considered to use `get_logger()` of node, but there is nowhere to involve ros2 node or no need to extend class `Cache` from ros2 node of which allows to use `get_logger()`, that's why I use the root logger instead, thanks !

src/message_filters/__init__.py
gaoethan: There is the following error which says 'zip' object is not subscriptable for test `test_approxsync.py`

```
12:             self.assertEqual(self.collector, [])
12:             for msg in random.sample(seq1, N):
12: >               m1.signalMessage(msg)
12:
12: test/test_approxsync.py:101:
12: _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
12: ../../build/message_filters/install/message_filters/lib/python3.6/site-packages/message_filters/__init__.py:63: in signalMessage
12:     cb(*(msg + args))
12: ../../build/message_filters/install/message_filters/lib/python3.6/site-packages/message_filters/__init__.py:295: in add
12:     for vv in itertools.product(*[zip(*s)[0] for s in stamps]):
12: _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
12:
12: .0 = <list_iterator object at 0x7f0dcde86908>
12:
12: >   for vv in itertools.product(*[zip(*s)[0] for s in stamps]):
12: E   TypeError: 'zip' object is not subscriptable
12:
12: ../../build/message_filters/install/message_filters/lib/python3.6/site-packages/message_filters/__init__.py:295: TypeError
12: =========================== 1 failed in 0.08 seconds ===========================
12: -- run_test.py: return code 1
12: -- run_test.py: verify result file '/home/ethan/Myspace/rosdev/2upstream/build/message_filters/build/message_filters/test_results/message_filters/test_approxsync.py.xunit.xml'
12/13 Test #12: test_approxsync.py .............................***Failed    0.94 sec

```

src/message_filters/__init__.py
gaoethan: Here `nanoseconds` is the [internal API](https://github.com/ros2/rclpy/blob/master/rclpy/rclpy/duration.py#L30-L32) of `Duration` , sorry, what does the `semantic datatypes` means from you here ? thanks !

src/message_filters/__init__.py
tfoote: I'm sorry I don't understand your statement that there's nowhere to involve a ros2 node, you have node `self.node` used on line 78 above. 

And as the `_root_logger` is prefixed by an underscore it's considered a private implementation detail you should not be directly calling it.

src/message_filters/__init__.py
tfoote: the `nanoseconds` call here returns the count of nanoseconds in the duration. It's a bare number. You should instead store the Duration object itself. The Duration object has the semantic meaning of representing a duration. Which can be added and subtracted to time or other duration objects, but can't be added to a distance. Whereas a bare number you can do anything you want with them and the implementer must keep track of the semantic meaning.

src/message_filters/__init__.py
tfoote: So this should store the time object. An example of how this might be abused would be trying to add two times which is not valid, but it is valid for durations.

src/message_filters/__init__.py
tfoote: Ahh apparently that's a python2 python3 change: https://stackoverflow.com/a/27431433/604099

src/message_filters/__init__.py
gaoethan: @tfoote the self.node used on line 78 above is in another `class Subscriber` but not for `class Cache`, so in order to use get_logger from a node within `class Cache`, it's required to pass a parameter of "ros2 node" into `class Cache` or have "class Cache extend from ros2 node class", but if the involving of ros2 node here is merely for logging, do you think it's really necessary ? thanks.

src/message_filters/__init__.py
gaoethan: Yes, it makes sense! But it will be problematic for the following (a.k.a line 255) case which can't be iterated and subscript for the semantic datatypes, anyway, let me try to optimize it ASAP. thanks.
```
 my_queue[stamp] = msg
```

tfoote:
For the short term please at least use the public API rclpy.logging.get_logger with a useful name. https://github.com/ros2/rclpy/blob/master/rclpy/rclpy/logging.py#L40

This will at least make it a little bit more readable.

However we've been looking at this elsewhere and for useful logging we will likely need to pass in the logger objects so that they can be remapped and otherwise grouped appropriately. So adding the logger to the init signature wouldn't be unreasonable. We can ticket that as a follow up.

gaoethan:
Now it's already changed to use the named logger with `rclpy.logging.get_logger()`. additionally, create the issue #8 for the follow-up optimization. thanks !

tfoote:
Pushed to https://github.com/ros2/message_filters/tree/refactor_py for easier CI running now:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5170/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1944/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4295/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5151/)

mikaelarguedas:
@tfoote I believe message_filters is not in the repos file so it will not be tested on regular CI.

gaoethan:
Link it to #2 for easy tracking

dirk-thomas:
@tfoote To run CI.

tfoote:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5223/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1984/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4342/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5200/)

mjcarroll:
This is the relevant failing bit on Windows:

```
03:24:38   C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.12.25827\bin\HostX86\x64\CL.exe /c /IC:\J\workspace\ci_windows\ws\src\ros\angles\include /IC:\J\workspace\ci_windows\ws\install\src\gtest_vendor\include /IC:\J\workspace\ci_windows\ws\install\include /nologo /W3 /WX- /diagnostics:classic /O2 /Ob2 /D WIN32 /D _WINDOWS /D NDEBUG /D "CMAKE_INTDIR=\"Release\"" /D _MBCS /Gm- /EHsc /MD /GS /fp:precise /Zc:wchar_t /Zc:forScope /Zc:inline /GR /std:c++14 /Fo"message_filters-test_exact_time_policy.dir\Release\\" /Fd"message_filters-test_exact_time_policy.dir\Release\vc141.pdb" /Gd /TP /errorReport:queue C:\J\workspace\ci_windows\ws\src\ros\angles\test\test_exact_time_policy.cpp
03:24:38   test_exact_time_policy.cpp
03:24:38 C:\J\workspace\ci_windows\ws\src\ros\angles\include\message_filters/sync_policies/exact_time.h(119): error C2059: syntax error: 'template' [C:\J\workspace\ci_windows\ws\build\message_filters\message_filters-test_exact_time_policy.vcxproj]
03:24:38   C:\J\workspace\ci_windows\ws\src\ros\angles\test\test_exact_time_policy.cpp(156): note: see reference to function template instantiation 'message_filters::Connection message_filters::sync_policies::ExactTime<Msg,Msg,message_filters::NullType,message_filters::NullType,message_filters::NullType,message_filters::NullType,message_filters::NullType,message_filters::NullType,message_filters::NullType>::registerDropCallback<std::_Binder<std::_Unforced,void (__cdecl Helper::* )(void),Helper *>>(const C &)' being compiled
03:24:38           with
03:24:38           [
03:24:38               C=std::_Binder<std::_Unforced,void (__cdecl Helper::* )(void),Helper *>
03:24:38           ]
03:24:38   C:\J\workspace\ci_windows\ws\src\ros\angles\test\test_exact_time_policy.cpp(156): note: see reference to function template instantiation 'message_filters::Connection message_filters::sync_policies::ExactTime<Msg,Msg,message_filters::NullType,message_filters::NullType,message_filters::NullType,message_filters::NullType,message_filters::NullType,message_filters::NullType,message_filters::NullType>::registerDropCallback<std::_Binder<std::_Unforced,void (__cdecl Helper::* )(void),Helper *>>(const C &)' being compiled
03:24:38           with
03:24:38           [
03:24:38               C=std::_Binder<std::_Unforced,void (__cdecl Helper::* )(void),Helper *>
03:24:38           ]
03:24:38 Done Building Project "C:\J\workspace\ci_windows\ws\build\message_filters\message_filters-test_exact_time_policy.vcxproj" (default targets) -- FAILED.
```

mikaelarguedas:
The failure looks C++ related and independent of these Python changes.
@tfoote @mjcarroll Was the initial port tested on Windows?
CI with [this additional repos file](https://gist.githubusercontent.com/mikaelarguedas/0394c84089fc9fb14ad3a1b2c84b6dea/raw/e721eef00edaba9f8734960e9a4be20dfcddecc5/message_filters.repos) to see if master builds:

[Build Status](https://ci.ros2.org/job/ci_windows/5204/)

tfoote:
@mjcarroll validated image_transport working on Windows so it was known to be working.

mjcarroll:
Image transport hasn't worked on CI on Windows, as it requires addition dependencies, but message_filters has worked in the past.  @tfoote Do you want to look into this or do you want me to?

tfoote:
We talked about this in the ROS2 meeting and since it's not a regression caused by this we'll merge this and ticket the regressions separately.

PR #14, opened 2018-11-16
Add the ability to use raw pointers in Subscriber

mjcarroll:
Using only SharedPtr was causing an issue in objects that are subclassed from Node.

If you want to create a `message_filter::Subscriber` as part of the subclass constructor, there was no way to get a shared pointer to the node instance itself to pass in. `shared_from_this` is not allowed in constructors for reasons listed here: https://en.cppreference.com/w/cpp/memory/enable_shared_from_this

As a workaround, I added an additional API where users can create a Subscriber using a raw pointer. The only downside to this is that the user is fully responsible for mananging the lifetime of the `message_filter::Subscriber` to ensure that it doesn't outlive the `Node`.  In the case where it is a member of a `Node` subclass, this is not an issue.

### DIFF
--- a/include/message_filters/subscriber.h
+++ b/include/message_filters/subscriber.h
@@ -46,17 +46,28 @@ namespace message_filters
 class SubscriberBase
 {
 public:
-  virtual ~SubscriberBase() {}
+  virtual ~SubscriberBase() = default;
   /**
    * \brief Subscribe to a topic.
    *
    * If this Subscriber is already subscribed to a topic, this function will first unsubscribe.
    *
-   * \param node The rclcpp::Node::ShardPtr to use to subscribe.
+   * \param node The rclcpp::Node::SharedPtr to use to subscribe.
    * \param topic The topic to subscribe to.
    * \param qos (optional) The rmw qos profile to use to subscribe
    */
   virtual void subscribe(rclcpp::Node::SharedPtr node, const std::string& topic, const rmw_qos_profile_t qos = rmw_qos_profile_default) = 0;
+
+  /**
+   * \brief Subscribe to a topic.
+   *
+   * If this Subscriber is already subscribed to a topic, this function will first unsubscribe.
+   *
+   * \param node The rclcpp::Node to use to subscribe.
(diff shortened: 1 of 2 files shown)


include/message_filters/subscriber.h
```diff
@@ -46,17 +46,28 @@ namespace message_filters
...
 {
 public:
-  virtual ~SubscriberBase() {}
+  virtual ~SubscriberBase() = default;
   /**
    * \brief Subscribe to a topic.
    *
    * If this Subscriber is already subscribed to a topic, this function will first unsubscribe.
    *
-   * \param node The rclcpp::Node::ShardPtr to use to subscribe.
```
wjwwood: Legit read this as ShardPlate... ⚔️

include/message_filters/subscriber.h
```diff
@@ -140,7 +154,43 @@ class Subscriber : public SubscriberBase, public SimpleFilter<M>
                [this](std::shared_ptr<M const> msg) {
                  this->cb(EventType(msg));
                }, qos);
-      node_ = node;
+
+      // The subscriber is switching to using a shared pointer, so relase the raw.
+      if (node_raw_ != nullptr) {
+        node_raw_ = nullptr;
+      }
+      node_shared_ = node;
```
wjwwood: It would be nice, to find a way to deduplicate the code here, maybe having the `shared_ptr` version call the raw pointer version and then change which is stored in `node_raw_` versus `node_shared_`.

test/test_subscriber.cpp
```diff
@@ -101,6 +166,7 @@ TEST(Subscriber, subUnsubSub)
   ASSERT_GT(h.count_, 0);
 }
 
+
```
wjwwood: nitpick: Minimize use of vertical whitespace, see: https://google.github.io/styleguide/cppguide.html#Vertical_Whitespace

include/message_filters/subscriber.h
mjcarroll: Please take a look at the updated logic.

PR #18, opened 2019-04-05
use Python debug interpreter on Windows

dirk-thomas:
Fixes the failing tests on Windows debug builds: https://ci.ros2.org/view/nightly/job/nightly_win_deb/1211/testReport/

[Build Status](https://ci.ros2.org/job/ci_windows/6455/)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -115,11 +115,24 @@ if(BUILD_TESTING)
     ament_target_dependencies(${PROJECT_NAME}-test_fuzz "rclcpp" "sensor_msgs")
   endif()
 
+  # Provides PYTHON_EXECUTABLE_DEBUG
+  find_package(python_cmake_module REQUIRED)
+  find_package(PythonExtra REQUIRED)
+  set(_PYTHON_EXECUTABLE "${PYTHON_EXECUTABLE}")
+  if(WIN32)
+    if(CMAKE_BUILD_TYPE STREQUAL "Debug")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -115,11 +115,24 @@ if(BUILD_TESTING)
     ament_target_dependencies(${PROJECT_NAME}-test_fuzz "rclcpp" "sensor_msgs")
   endif()
 
+  # Provides PYTHON_EXECUTABLE_DEBUG
+  find_package(python_cmake_module REQUIRED)
```
nuclearsandwich: Does a dependency on python_cmake_module need to be added to the `package.xml`?

dirk-thomas: The dependency was already used before. In fact it is already `find_package()`-ed above: https://github.com/ros2/message_filters/blob/9b95dfa4bdaa2893f4f8ad5ec722b0ad53071690/CMakeLists.txt#L45 Therefore I removed the duplicate call in bbe651c25acf0820e899d587c64295038715d2cb.

Since it is unrelated to this fix I don't plan to change it. I will leave it up to the maintainer / whoever has ported the ported the package to change the manifest if they think it should be done.

nuclearsandwich: Opened https://github.com/ros2/message_filters/pull/19 since I believe shared custodianship and fixing small things when they're noticed produces better software in the long term.

PR #24, opened 2019-05-01
Make format string agree with argument type.

nuclearsandwich:
This resolves warnings which popped up as a result of
https://github.com/ros2/rcutils/pull/154.

### DIFF
--- a/include/message_filters/sync_policies/approximate_time.h
+++ b/include/message_filters/sync_policies/approximate_time.h
@@ -194,8 +194,8 @@ struct ApproximateTime : public PolicyBase<M0, M1, M2, M3, M4, M5, M6, M7, M8>
     else if ((msg_time - previous_msg_time) < inter_message_lower_bounds_[i])
     {
       RCUTILS_LOG_WARN_ONCE("Messages of type %d arrived closer ("
-        "%g ) than the lower bound you provided ("
-        "%g) (will print only once)",
+        "%ld ) than the lower bound you provided ("
+        "%ld) (will print only once)",
         i,
         (msg_time - previous_msg_time).nanoseconds(),
         inter_message_lower_bounds_[i].nanoseconds());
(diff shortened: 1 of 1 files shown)


include/message_filters/sync_policies/approximate_time.h
```diff
@@ -194,8 +194,8 @@ struct ApproximateTime : public PolicyBase<M0, M1, M2, M3, M4, M5, M6, M7, M8>
     else if ((msg_time - previous_msg_time) < inter_message_lower_bounds_[i])
     {
       RCUTILS_LOG_WARN_ONCE("Messages of type %d arrived closer ("
-        "%g ) than the lower bound you provided ("
-        "%g) (will print only once)",
+        "%ld ) than the lower bound you provided ("
+        "%ld) (will print only once)",
```
dirk-thomas: `PRId64` looks like the right thing for `int64_t`.

include/message_filters/sync_policies/approximate_time.h
```diff
@@ -194,8 +194,8 @@ struct ApproximateTime : public PolicyBase<M0, M1, M2, M3, M4, M5, M6, M7, M8>
     else if ((msg_time - previous_msg_time) < inter_message_lower_bounds_[i])
     {
       RCUTILS_LOG_WARN_ONCE("Messages of type %d arrived closer ("
-        "%g ) than the lower bound you provided ("
-        "%g) (will print only once)",
+        "%ld ) than the lower bound you provided ("
```
clalancette: From what I can tell, this is printing out a `rclcpp::Time.nanoseconds()`, which is an int64_t.  While the above most likely works on 64-bit arches, the truly portable way to do this is with `%PRId64` (that way it won't warn on 32-bit either).  I'm going to change these two to use that instead.

clalancette:
CI is green:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6873/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3159/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5646/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6693/)

PR #27, opened 2019-05-14
Fix deprecation warning

jacobperron:
Connected to https://github.com/ros2/rclpy/issues/342


jacobperron:
> Probably didn't mean to check in those pycache files.

Oops, removed. :grimacing:

mjcarroll:
Perhaps an update to .gitignore?

jacobperron:
There was no .gitignore, so I copied the one from `rclpy`.

PR #38, opened 2019-08-31
Add policy to store the latest messages

jcmonteiro:
Add policies to store the most recent received messages (based on their timestamp or on the order of arrival); These policies **do not** call any callback automatically. It is up to the programmer to call the `call` method when it is time to process the information.

These policies are meant to be used in networks where information is processed periodically.

### DIFF
--- a/include/message_filters/sync_policies/latest_stamped.h
+++ b/include/message_filters/sync_policies/latest_stamped.h
@@ -1,0 +1,142 @@
+/*********************************************************************
+* Software License Agreement (BSD License)
+*
+*  Copyright (c) 2009, Willow Garage, Inc.
+*  All rights reserved.
+*
+*  Redistribution and use in source and binary forms, with or without
+*  modification, are permitted provided that the following conditions
+*  are met:
+*
+*   * Redistributions of source code must retain the above copyright
+*     notice, this list of conditions and the following disclaimer.
+*   * Redistributions in binary form must reproduce the above
+*     copyright notice, this list of conditions and the following
+*     disclaimer in the documentation and/or other materials provided
+*     with the distribution.
+*   * Neither the name of the Willow Garage nor the names of its
+*     contributors may be used to endorse or promote products derived
+*     from this software without specific prior written permission.
+*
+*  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+*  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+*  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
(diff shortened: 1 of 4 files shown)


ZhenshengLee:
> It is up to the programmer to call the `call` method when it is time to process the information.

@jcmonteiro 

Could you provide some docs to explain the motivation of this policy? or more info about why, when and how we should use this policy?

include/message_filters/sync_policies/latest_stamped.h
```diff
@@ -0,0 +1,142 @@
...
+*  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+*  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+*  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+*  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+*  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+*  POSSIBILITY OF SUCH DAMAGE.
+*********************************************************************/
+
+#ifndef MESSAGE_FILTERS__SYNC_LatestStamped_H_
+#define MESSAGE_FILTERS__SYNC_LatestStamped_H_
```
clalancette, line 36:
```suggestion
#ifndef MESSAGE_FILTERS__SYNC_LATESTSTAMPED_H_
#define MESSAGE_FILTERS__SYNC_LATESTSTAMPED_H_
```

clalancette (changes requested):
I know this is really old, so sorry for the long delay in looking at it.  If you'd like to give up on getting this in, I'd totally understand.

If you'd still like to move forward, I've left some stuff inline to fix.  Also, I'm not a huge fan of the terminology "latest" to describe this policy.  While it is accurate, I don't think it will be hugely meaningful to people, and seems too close to "latest_time" (which we already have).  Maybe we can rename it to "user controlled", or something like that.

ZhenshengLee:
@jcmonteiro  Friendly ping for updates.

ZhenshengLee:
@jcmonteiro Would you agree that I continue to work with your branch and complete with the PR? 

I'll begin with following the change requests in this PR, then create a new PR with the forked branch with my commits.

jcmonteiro:
> @jcmonteiro Would you agree that I continue to work with your branch and complete with the PR?
> 
> I'll begin with following the change requests in this PR, then create a new PR with the forked branch with my commits.

Hi @ZhenshengLee. I've been out of the loop of the robotics and ROS communities for a while now. Please feel free to take this PR in the direction that you see fit.

ZhenshengLee:
Close this and continue with #107

Issue #39, opened 2019-10-08
message filter doesn't allow non-default quality of service


Jconn:
The subscription created in the Subscriber `__init__` has a hard-coded queue depth of 10. If `qos_profile` is set as a named param like I am doing:

```
        self.pc_filter = message_filters.Subscriber(
                self,
                PointCloud2,
                constants.PC2_TOPIC,
                qos_profile = qos_profile_sensor_data
                )

```
Then an exception is thrown, because qos_profile is being set twice.
I am throwing around a lot of pointcloud data, and I really want to be able to specify a best effort quality of service For some of my filters. 

Thank you for the clear type annotations, making this easy to track down.

for reference, the problem `__init__` function is shown below:
```
class Subscriber(SimpleFilter):
    def __init__(self, *args, **kwargs):
        SimpleFilter.__init__(self)
        self.node = args[0]
        self.topic = args[2]
        self.sub = self.node.create_subscription(*args[1:], self.callback, 10, **kwargs)
```

wjwwood:
@claireyywang can you close this or update it as needed?

claireyywang:
> @claireyywang can you close this or update it as needed?

Yeah sure my bad. Closing as the PR is merged.

PR #41, opened 2019-10-24
allow custom qos for message filters

Jconn:
this fixes issue https://github.com/ros2/message_filters/issues/39 and allows me to explicitly set QoS profiles for each of my message filter subscribers

### DIFF
--- a/src/message_filters/__init__.py
+++ b/src/message_filters/__init__.py
@@ -79,7 +79,7 @@ class Subscriber(SimpleFilter):
         if kwargs.get('qos_profile'):
             self.sub = self.node.create_subscription(*args[1:], self.callback, **kwargs)
         else:
-            self.sub = self.node.create_subscription(*args[1:], self.callback, **kwargs)
+            self.sub = self.node.create_subscription(*args[1:], self.callback, 10, **kwargs)
 
     def callback(self, msg):
         self.signalMessage(msg)
--- a/src/message_filters/__init__.py
+++ b/src/message_filters/__init__.py
@@ -76,7 +76,10 @@ class Subscriber(SimpleFilter):
         SimpleFilter.__init__(self)
         self.node = args[0]
         self.topic = args[2]
-        self.sub = self.node.create_subscription(*args[1:], self.callback, **kwargs)
+        if kwargs.get('qos_profile'):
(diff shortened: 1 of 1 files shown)


claireyywang:
I added a check to prevent breaking API for current users who are not supplying their own `qos_profile` params. @ros2/team Should we deprecate the hardcoded default `qos_profile` and just ask users to always supply their own params in the future?

src/message_filters/__init__.py
```diff
@@ -76,7 +76,10 @@ def __init__(self, *args, **kwargs):
         SimpleFilter.__init__(self)
         self.node = args[0]
         self.topic = args[2]
-        self.sub = self.node.create_subscription(*args[1:], self.callback, 10, **kwargs)
+        if kwargs.get('qos_profile'):
+            self.sub = self.node.create_subscription(*args[1:], self.callback, **kwargs)
+        else:
+            self.sub = self.node.create_subscription(*args[1:], self.callback, 10, **kwargs)
```
wjwwood: You could mutate the `kwargs` rather than having `if`/`else` statements. Or at least, if you're going to do it this way, use the keyword before the `10`, I think it would be `qos_profile=10`.

jacobperron: Agreed. I would do something like `kwargs.setdefault('qos_profile', 10)`.

wjwwood:
> I don't feel strongly about adding a deprecation warning. The other option is to just hard break the API post Eloquent.

I think that's the big question we need feedback on. It would be nice to have a deprecation cycle on this, and to get that deprecation in for Eloquent.

@mjcarroll I think you poked around on message filters too, do you have an opinion?

jacobperron:
I agree that it would be nice to have a deprecation cycle, which should be straightforward with the Python API. It would be nice to also make the same change to the C++ implementation (there's a TODO!):

https://github.com/ros2/message_filters/blob/01e433782c3e395230016138d3d1fb91553d57c9/include/message_filters/subscriber.h#L161-L162

Maybe it's as easy as overloading that method and deprecating the old one.

claireyywang:
@wjwwood @mjcarroll friendly ping. If we are okay with the current change, could I get a green check to merge? Maybe we can discuss deprecation cycle in meeting.

PR #46, opened 2020-06-02
export targets in addition to include directories / libraries

dirk-thomas:
Necessary to make downstream packages like `rviz_common` which already use modern CMake targets relocatable.

See ros2/rviz#556.


dirk-thomas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10978/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6315/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8938/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10883/)

PR #50, opened 2020-12-04
Add Readme with example of python usage

charlielito:
Added a minimal working example and caveats for the python api after  having a bad time trying to look info on the internet (without success).
TODO for c++ users: Add c++ api and example

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,45 @@
+# ROS2 message filters
+Package that implements different filters for ros messages.
+
+## C++ api
+
+## Python Api
+The python api has the following classes:
+* Subscriber: ROS2 subscription filter, similar arguments as `rclpy.Subscriber`.
+* Cache: Stores a time history of messages.
+* TimeSynchronizer: Synchronizes messages by their timestamps.
+* ApproximateTimeSynchronizer: Approximately synchronizes messages by their timestamps.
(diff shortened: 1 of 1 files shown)


clalancette (commented):
So, there already sort of is this content in the `index.rst` file.  It doesn't have quite all of the parts of this example, but it does have some better explanations of what the various filter types are.  I'm wondering if we should just slightly expand the index.rst, and then link this README to that (so we don't duplicate content).

charlielito:
My bad, I didn't notice the `index.rst` before :sweat: 
Yes, that could be a nice way to go, modifying the `index.rst` and just leave a simple Readme with a link to that file, for blind people like me :smile: 
Also I think the example given in `index.rst` does not work:
```python
from message_filters import TimeSynchronizer, Subscriber

def gotimage(image, camerainfo):
    assert image.header.stamp == camerainfo.header.stamp
    print "got an Image and CameraInfo"

tss = TimeSynchronizer(Subscriber("/wide_stereo/left/image_rect_color", sensor_msgs.msg.Image),
                       Subscriber("/wide_stereo/left/camera_info", sensor_msgs.msg.CameraInfo))
tss.registerCallback(gotimage)
```
`message_filters.Subscriber` expects as first argument a ros2 node object, so this won't run

clalancette:
Based on the previous conversation, closing this out.  We should do a separate PR which updates index.rst, and also adds a link from the README to that file.

PR #51, opened 2021-01-31
Add rclcpp::QoS constructors and methods

fmrico:
Hi @wjwwood 

As you suggested, I have upgraded to the new API.

Please, review it for any issue
Francisco

Signed-off-by: Francisco Martin Rico <fmrico@gmail.com>

### DIFF
--- a/include/message_filters/subscriber.h
+++ b/include/message_filters/subscriber.h
@@ -57,6 +57,16 @@ public:
    * \param qos (optional) The rmw qos profile to use to subscribe
    */
   virtual void subscribe(rclcpp::Node::SharedPtr node, const std::string& topic, const rmw_qos_profile_t qos = rmw_qos_profile_default) = 0;
+  /**
+   * \brief Subscribe to a topic.
+   *
+   * If this Subscriber is already subscribed to a topic, this function will first unsubscribe.
+   *
+   * \param node The rclcpp::Node::SharedPtr to use to subscribe.
+   * \param topic The topic to subscribe to.
+   * \param qos (optional) The qos profile to use to subscribe
+   */
+  virtual void subscribe(rclcpp::Node::SharedPtr node, const std::string& topic, const rclcpp::QoS & qos) = 0;
 
   /**
    * \brief Subscribe to a topic.
@@ -67,7 +77,17 @@ public:
    * \param topic The topic to subscribe to.
    * \param qos (optional) The rmw qos profile to use to subscribe
    */
-  virtual void subscribe(rclcpp::Node * node, const std::string& topic, const rmw_qos_profile_t qos = rmw_qos_profile_default) = 0;
+  virtual void subscribe(rclcpp::Node * node, const std::string& topic, const rmw_qos_profile_t qos= rmw_qos_profile_default) = 0;
+  /**
+   * \brief Subscribe to a topic.
+   *
+   * If this Subscriber is already subscribed to a topic, this function will first unsubscribe.
+   *
+   * \param node The rclcpp::Node to use to subscribe.
+   * \param topic The topic to subscribe to.
+   * \param qos (optional) The qos profile to use to subscribe
+   */
+  virtual void subscribe(rclcpp::Node * node, const std::string& topic, const rclcpp::QoS & qos) = 0;
   /**
    * \brief Re-subscribe to a topic.  Only works if this subscriber has previously been subscribed to a topic.
    */
@@ -113,12 +133,33 @@ public:
    * \param topic The topic to subscribe to.
    * \param qos (optional) The rmw qos profile to use to subscribe
    */
-  Subscriber(rclcpp::Node::SharedPtr node, const std::string& topic, const rmw_qos_profile_t qos = rmw_qos_profile_default)
+  Subscriber(rclcpp::Node::SharedPtr node, const std::string& topic, const rmw_qos_profile_t qos = rmw_qos_profile_default) 
+  {
+    rclcpp::QoS rclcpp_qos(rclcpp::QoSInitialization::from_rmw(qos));
+    subscribe(node, topic, rclcpp_qos);
+  }
+
+  Subscriber(rclcpp::Node* node, const std::string& topic, const rmw_qos_profile_t qos = rmw_qos_profile_default)
+  {
+    rclcpp::QoS rclcpp_qos(rclcpp::QoSInitialization::from_rmw(qos));
+    subscribe(node, topic, rclcpp_qos);
+  }
+
+  /**
+   * \brief Constructor
+   *
+   * See the rclcpp::Node::subscribe() variants for more information on the parameters
+   *
+   * \param node The rclcpp::Node::SharedPtr to use to subscribe.
+   * \param topic The topic to subscribe to.
+   * \param qos (optional) The qos profile to use to subscribe
+   */
+  Subscriber(rclcpp::Node::SharedPtr node, const std::string& topic, const rclcpp::QoS & qos)
   {
     subscribe(node, topic, qos);
   }
 
-  Subscriber(rclcpp::Node* node, const std::string& topic, const rmw_qos_profile_t qos = rmw_qos_profile_default)
+  Subscriber(rclcpp::Node* node, const std::string& topic, const rclcpp::QoS & qos)
   {
     subscribe(node, topic, qos);
   }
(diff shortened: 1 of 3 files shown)


include/message_filters/subscriber.h
```diff
@@ -113,12 +133,33 @@ class Subscriber : public SubscriberBase, public SimpleFilter<M>
    * \param topic The topic to subscribe to.
    * \param qos (optional) The rmw qos profile to use to subscribe
    */
-  Subscriber(rclcpp::Node::SharedPtr node, const std::string& topic, const rmw_qos_profile_t qos = rmw_qos_profile_default)
+  Subscriber(rclcpp::Node::SharedPtr node, const std::string& topic, const rmw_qos_profile_t qos = rmw_qos_profile_default)
```
wjwwood, line 136: Probably unintentional trailing whitespace.

```suggestion
  Subscriber(rclcpp::Node::SharedPtr node, const std::string& topic, const rmw_qos_profile_t qos = rmw_qos_profile_default)
```

include/message_filters/subscriber.h
```diff
@@ -67,7 +77,17 @@ class SubscriberBase
    * \param topic The topic to subscribe to.
    * \param qos (optional) The rmw qos profile to use to subscribe
    */
-  virtual void subscribe(rclcpp::Node * node, const std::string& topic, const rmw_qos_profile_t qos = rmw_qos_profile_default) = 0;
+  virtual void subscribe(rclcpp::Node * node, const std::string& topic, const rmw_qos_profile_t qos= rmw_qos_profile_default) = 0;
```
wjwwood, line 80:
```suggestion
  virtual void subscribe(rclcpp::Node * node, const std::string& topic, const rmw_qos_profile_t qos = rmw_qos_profile_default) = 0;
```

include/message_filters/subscriber.h
```diff
@@ -140,9 +181,24 @@ class Subscriber : public SubscriberBase, public SimpleFilter<M>
    *
    * \param node The rclcpp::Node::SharedPtr to use to subscribe.
    * \param topic The topic to subscribe to.
-   * \param qos (optional) The rmw qos profile to use to subscribe
+   * \param qos (optional) The qos profile to use to subscribe
```
wjwwood, line 184: Not sure if this is better or worse, this signature is still using the rmw version.

include/message_filters/subscriber.h
```diff
@@ -140,9 +181,24 @@ class Subscriber : public SubscriberBase, public SimpleFilter<M>
...
+  }
+
+  /**
+   * \brief Subscribe to a topic.
+   *
+   * If this Subscriber is already subscribed to a topic, this function will first unsubscribe.
+   *
+   * \param node The rclcpp::Node::SharedPtr to use to subscribe.
+   * \param topic The topic to subscribe to.
+   * \param qos (optional) The qos profile to use to subscribe
```
wjwwood, line 199: According to the signature it is not optional.

```suggestion
   * \param qos The qos profile to use to subscribe
```

include/message_filters/subscriber.h
```diff
@@ -156,20 +212,32 @@ class Subscriber : public SubscriberBase, public SimpleFilter<M>
    *
    * \param node The rclcpp::Node to use to subscribe.
    * \param topic The topic to subscribe to.
-   * \param qos (optional) The rmw qos profile to use to subscribe
+   * \param qos (optional) The qos profile to use to subscribe
    */
-  // TODO(wjwwood): deprecate in favor of API's that use `rclcpp::QoS` instead.
```
wjwwood, line 161: In order to remove this, I'd recommend actually deprecating the old version in favor of the new one that requires the QoS be specified.

include/message_filters/subscriber.h
```diff
@@ -241,7 +309,7 @@ class Subscriber : public SubscriberBase, public SimpleFilter<M>
   rclcpp::Node * node_raw_ {nullptr};
 
   std::string topic_;
-  rmw_qos_profile_t qos_;
+  rclcpp::QoS qos_ {100};
```
wjwwood, line 312: Maybe this should be a pointer rather than arbitrarily initialized to this.

`qos_` is only used in the `subscribe()` method which "resubscribes", but I wonder what should happen if that method is called after construction but before an explicit subscribe with arguments is made?

Some of the constructors call subscribe with arguments, so maybe this should be initialized in the constructor instead?

Either way, this arbitrary default doesn't make much sense to me.

include/message_filters/subscriber.h
fmrico: Yes, you are right. I would have removed the old API, but I suppose there is much code depending on this.  I have added a message to the rmw methods: ` [[deprecated("Use rclcpp::QoS to specify qos")]]`

fmrico:
Hi @wjwwood 

I have also added support for rclcpp_lifecycle. I can move it to another PR, but if you find it adequate here, I could include it, adding some extra tests.

Thanks!

include/message_filters/subscriber.h
fmrico: Fixed

include/message_filters/subscriber.h
fmrico: Fixed

include/message_filters/subscriber.h
fmrico: I mantained

include/message_filters/subscriber.h
fmrico: Fixed

include/message_filters/subscriber.h
fmrico: I have removed this and initialized it in constructors

fmrico:
Hi @wjwwood 

Any review/comment to continue with this?

include/message_filters/subscriber.h
```diff
@@ -52,22 +53,34 @@ class SubscriberBase
...
    * If this Subscriber is already subscribed to a topic, this function will first unsubscribe.
    *
-   * \param node The rclcpp::Node::SharedPtr to use to subscribe.
+   * \param node The rclcpp::Node::SharedPtr or rclcpp_lifecycle::LifeCycleNode::SharedPtr  to use to subscribe.
    * \param topic The topic to subscribe to.
    * \param qos (optional) The rmw qos profile to use to subscribe
    */
-  virtual void subscribe(rclcpp::Node::SharedPtr node, const std::string& topic, const rmw_qos_profile_t qos = rmw_qos_profile_default) = 0;
-
+  // [[deprecated("Use rclcpp::QoS to specify qos")]]
```
wjwwood, line 60: Sorry it took so long for me to get back to this, but why are these commented out now?

package.xml
```diff
@@ -20,6 +20,7 @@
 
   <build_depend>builtin_interfaces</build_depend>
   <build_depend>rclcpp</build_depend>
+  <build_depend>rclcpp_lifecycle</build_depend>
```
wjwwood, line 23: Why do you need to depend on `rclcpp_lifecycle` here?

I think you don't use it at all, other than to reference it in the documentation.

include/message_filters/subscriber.h
```diff
@@ -52,22 +53,34 @@ class SubscriberBase
    *
    * If this Subscriber is already subscribed to a topic, this function will first unsubscribe.
    *
-   * \param node The rclcpp::Node::SharedPtr to use to subscribe.
+   * \param node The rclcpp::Node::SharedPtr or rclcpp_lifecycle::LifeCycleNode::SharedPtr  to use to subscribe.
```
wjwwood, line 56: This is kind of inaccurate, actually, it's really more like:

```suggestion
   * \param node Shared pointer to the rclcpp::Node::SharedPtr object, or to an object
   *   that fulfills the required NodeT interface, to use to subscribe.
```

It need not be just `rclcpp::Node` or `rclcpp_lifecycle::LifeCycleNode`, it could be other things as well.

include/message_filters/subscriber.h
```diff
@@ -43,6 +43,7 @@
 namespace message_filters
 {
 
+template <class NodeT = rclcpp::Node>
```
wjwwood, line 46: If you stored pointers to the various Node interface functions you needed, you could avoid making this class templated, and just template the methods.

See this function:

https://github.com/ros2/rclcpp/blob/98fc7fecc25b2fecc089d1117cee6803eda34836/rclcpp/include/rclcpp/create_subscription.hpp#L205-L235

You could store just the `rclcpp::node_interfaces::NodeParametersInterface::SharedPtr` and `rclcpp::node_interfaces::NodeTopicsInterface::SharedPtr` from the `NodeT` and then call the above function to subscribe to things.

clalancette:
@fmrico Friendly ping here, can you respond to @wjwwood 's review?

clalancette:
I'm going to close this one out for now.  If you'd like to continue it in the future, please feel free to reopen and respond to William's comments.  Thanks for the contribution.

PR #53, opened 2021-04-28
Expose subscription options

audrow:
This PR is needed for QoS overrides in subscriptions, as in https://github.com/ros-perception/image_pipeline/pull/651.

### DIFF
--- a/include/message_filters/subscriber.h
+++ b/include/message_filters/subscriber.h
@@ -43,6 +43,7 @@
 namespace message_filters
 {
 
+template <typename AllocatorT=std::allocator<void>>
 class SubscriberBase
 {
 public:
@@ -56,7 +57,12 @@ public:
    * \param topic The topic to subscribe to.
    * \param qos (optional) The rmw qos profile to use to subscribe
    */
-  virtual void subscribe(rclcpp::Node::SharedPtr node, const std::string& topic, const rmw_qos_profile_t qos = rmw_qos_profile_default) = 0;
+  virtual void subscribe(
+    rclcpp::Node::SharedPtr node,
+    const std::string& topic,
+    const rmw_qos_profile_t qos = rmw_qos_profile_default,
+    rclcpp::SubscriptionOptionsWithAllocator<AllocatorT> options =
+      rclcpp::SubscriptionOptionsWithAllocator<AllocatorT>()) = 0;
 
   /**
    * \brief Subscribe to a topic.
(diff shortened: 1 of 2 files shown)


test/test_subscriber.cpp
```diff
@@ -63,7 +66,7 @@ TEST(Subscriber, simple)
 {
   auto node = std::make_shared<rclcpp::Node>("test_node");
   Helper h;
-  Subscriber<Msg> sub(node, "test_topic");
+  Subscriber<Msg> sub(node, "test_topic", g_qos_profile, g_options);
```
jacobperron, line 69: Are these changes to the tests necessary? I figured that it shouldn't be given the default parameters.

include/message_filters/subscriber.h
```diff
@@ -56,7 +57,12 @@ class SubscriberBase
    * \param topic The topic to subscribe to.
    * \param qos (optional) The rmw qos profile to use to subscribe
    */
-  virtual void subscribe(rclcpp::Node::SharedPtr node, const std::string& topic, const rmw_qos_profile_t qos = rmw_qos_profile_default) = 0;
+  virtual void subscribe(
+    rclcpp::Node::SharedPtr node,
+    const std::string& topic,
+    const rmw_qos_profile_t qos = rmw_qos_profile_default,
+    rclcpp::SubscriptionOptionsWithAllocator<AllocatorT> options =
+      rclcpp::SubscriptionOptionsWithAllocator<AllocatorT>()) = 0;
```
jacobperron, line 65: It would be nice if we could introduce these options without breaking API, but I'm not sure if that's possible in a pure virtual class.

To be honest, it's not clear to me why we have this abstract class to begin with, it doesn't appear to be used anywhere (besides being implemented by `Subscriber` below) :thinking:  It might be safer to omit the changes to `SubscriberBase` and add new overloads to `Subscriber` to incorporate the subscriber options.

Pinging @mjcarroll for thoughts (I think you're the maintainer?).

test/test_subscriber.cpp
audrow: I think you're right: they're not necessary given default parameters.

include/message_filters/subscriber.h
ivanpauno: > To be honest, it's not clear to me why we have this abstract class to begin with

It doesn't seem to be meant to be extended by external users, so it might be ok to just add new virtual methods to it.

jacobperron: If it's not meant to be extended, then I think we could just add the new method to the concrete class. 

We could go so far as to deprecate the pure virtual class and/or remove it in a separate PR, to avoid similar API/ABI issues in the future.

jacobperron: I'm also okay with your proposal to implement and warn about the non-overridden method https://github.com/ros2/message_filters/pull/53#discussion_r626682459

audrow:
Closed in favor of #56.

PR #56, opened 2021-05-17
Expose Subscription Options - V2

audrow:
This PR is needed for QoS overrides in subscriptions, as in https://github.com/ros-perception/image_pipeline/pull/651.

This is the approach suggested in https://github.com/ros2/message_filters/pull/53#discussion_r626682459.

### DIFF
--- a/include/message_filters/subscriber.h
+++ b/include/message_filters/subscriber.h
@@ -107,7 +107,9 @@ public:
     rclcpp::SubscriptionOptions options)
   {
     (void) options;
-    RCLCPP_WARN(node->get_logger(), "SubscriberBase::subscribe with four arguments has not been overridden");
+    RCLCPP_ERROR(
+      node->get_logger(),
+      "SubscriberBase::subscribe with four arguments has not been overridden");
     this->subscribe(node, topic, qos);
   }
 
--- a/include/message_filters/subscriber.h
+++ b/include/message_filters/subscriber.h
@@ -35,6 +35,8 @@
 #ifndef MESSAGE_FILTERS__SUBSCRIBER_H_
 #define MESSAGE_FILTERS__SUBSCRIBER_H_
 
+#include <stdexcept>
+
 #include <rclcpp/rclcpp.hpp>
 
 #include "message_filters/connection.h"
@@ -68,6 +70,47 @@ public:
    * \param qos (optional) The rmw qos profile to use to subscribe
    */
   virtual void subscribe(rclcpp::Node * node, const std::string& topic, const rmw_qos_profile_t qos = rmw_qos_profile_default) = 0;
+
+  /**
+   * \brief Subscribe to a topic.
+   *
+   * If this Subscriber is already subscribed to a topic, this function will first unsubscribe.
+   * This override allows SubscriptionOptions to be passed into the class without changing API.
+   *
+   * \param node The rclcpp::Node::SharedPtr to use to subscribe.
+   * \param topic The topic to subscribe to.
+   * \param qos (optional) The rmw qos profile to use to subscribe.
+   * \param options The subscription options to use to subscribe.
+   */
+  virtual void subscribe(
+    rclcpp::Node::SharedPtr node,
+    const std::string& topic,
+    const rmw_qos_profile_t qos,
+    rclcpp::SubscriptionOptions options)
+  {
+    this->subscribe(node.get(), topic, qos, options);
+  };
+
+  /**
+   * \brief Subscribe to a topic.
+   *
+   * If this Subscriber is already subscribed to a topic, this function will first unsubscribe.
+   *
+   * \param node The rclcpp::Node to use to subscribe.
+   * \param topic The topic to subscribe to.
+   * \param qos (optional) The rmw qos profile to use to subscribe
+   */
+  virtual void subscribe(
+    rclcpp::Node * node,
+    const std::string& topic,
+    const rmw_qos_profile_t qos,
+    rclcpp::SubscriptionOptions options)
+  {
+    (void) options;
+    throw std::runtime_error("SubscriberBase::subscribe with four arguments has not been overridden");
+    this->subscribe(node, topic, qos);
+  }
+
   /**
    * \brief Re-subscribe to a topic.  Only works if this subscriber has previously been subscribed to a topic.
    */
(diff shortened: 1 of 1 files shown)


include/message_filters/subscriber.h
```diff
@@ -68,6 +70,47 @@ class SubscriberBase
...
+   * \param qos (optional) The rmw qos profile to use to subscribe
+   */
+  virtual void subscribe(
+    rclcpp::Node * node,
+    const std::string& topic,
+    const rmw_qos_profile_t qos,
+    rclcpp::SubscriptionOptions options)
+  {
+    (void) options;
+    throw std::runtime_error("SubscriberBase::subscribe with four arguments has not been overridden");
```
jacobperron, line 110: I think we want to log a warning here, not throw. If we throw, then the code on L111 is never executed.
We could use the node's logger (I haven't tested this code):

```c++
RCLCPP_WARN(node->get_logger(), "SubscriberBase::subscribe with four arguments has not been overridden");
```

include/message_filters/subscriber.h
```diff
@@ -68,6 +70,47 @@ class SubscriberBase
...
+   *
+   * \param node The rclcpp::Node::SharedPtr to use to subscribe.
+   * \param topic The topic to subscribe to.
+   * \param qos (optional) The rmw qos profile to use to subscribe.
+   * \param options The subscription options to use to subscribe.
+   */
+  virtual void subscribe(
+    rclcpp::Node::SharedPtr node,
+    const std::string& topic,
+    const rmw_qos_profile_t qos,
```
ivanpauno, line 88: Feel free to ignore

----

I would take advantake and use `rclcpp::QoS` in the new methods.

You can use `qos.get_rmw_qos_profile()` in the default implementation to call the old method.

include/message_filters/subscriber.h
```diff
@@ -68,6 +70,47 @@ class SubscriberBase
...
+   * If this Subscriber is already subscribed to a topic, this function will first unsubscribe.
+   *
+   * \param node The rclcpp::Node to use to subscribe.
+   * \param topic The topic to subscribe to.
+   * \param qos (optional) The rmw qos profile to use to subscribe
+   */
+  virtual void subscribe(
+    rclcpp::Node * node,
+    const std::string& topic,
+    const rmw_qos_profile_t qos,
```
ivanpauno, line 106: same here

include/message_filters/subscriber.h
```diff
@@ -68,6 +70,47 @@ class SubscriberBase
...
+   * If this Subscriber is already subscribed to a topic, this function will first unsubscribe.
+   * This override allows SubscriptionOptions to be passed into the class without changing API.
+   *
+   * \param node The rclcpp::Node::SharedPtr to use to subscribe.
+   * \param topic The topic to subscribe to.
+   * \param qos (optional) The rmw qos profile to use to subscribe.
+   * \param options The subscription options to use to subscribe.
+   */
+  virtual void subscribe(
+    rclcpp::Node::SharedPtr node,
```
ivanpauno, line 86: I'm really surprised of how the current code accepts both a `Node::SharedPtr` and a `Node *` :man_facepalming:, but it doesn't matter for the purposes of this PR.

ivanpauno (approved):

include/message_filters/subscriber.h
jacobperron: Yeah, it may make sense to refactor the argument to be a "node-like" type by making it a template. Similar to how we do it in other places. But this is probably best left to a future PR.

include/message_filters/subscriber.h
```diff
@@ -68,6 +70,47 @@ class SubscriberBase
...
+   * \param qos (optional) The rmw qos profile to use to subscribe
+   */
+  virtual void subscribe(
+    rclcpp::Node * node,
+    const std::string& topic,
+    const rmw_qos_profile_t qos,
+    rclcpp::SubscriptionOptions options)
+  {
+    (void) options;
+    RCLCPP_WARN(node->get_logger(), "SubscriberBase::subscribe with four arguments has not been overridden");
```
jacobperron, line 110: On second thought, I think this should be `RCLCPP_ERROR`, since it's clearly a bug if someone calls this and the `options` are ignored. Sorry for changing my mind...

audrow: No worries

include/message_filters/subscriber.h
audrow: My feeling is to do this in a separate PR.

include/message_filters/subscriber.h
ivanpauno: > Yeah, it may make sense to refactor the argument to be a "node-like" type by making it a template. Similar to how we do it in other places. But this is probably best left to a future PR.

I wasn't exactly referring to that (a template would be a problem in a virtual interface).
Currently the `Subscriber` implementation accepts both storing a non-owning pointer to the node or a pointer to the node with shared ownership: https://github.com/ros2/message_filters/blob/667f1d83cc0faaf95d674db0b56c9187b4ccf0e8/include/message_filters/subscriber.h#L353-L354

That's extremely confusing, because in some cases the `message_filters::Subscriber` has ownership of the `Node` and in others doesn't, having the user to make sure that the node is still alive.

We also sometimes store the `rclcpp::Node*Interface` (e.g. `rclcpp::NodeParametersInterface`) directly, instead of a `Node` pointer.

We should really have recommendations of how to do this, and also review all the ownership relations between objects in `rclcpp` (I have seen way too many use-after-free bug reports already...)

include/message_filters/subscriber.h
ivanpauno: sounds good to me

ivanpauno (approved):
LGTM with green CI

jacobperron (approved):

include/message_filters/subscriber.h
jacobperron: That is confusing.

> We also sometimes store the rclcpp::Node*Interface (e.g. rclcpp::NodeParametersInterface) directly, instead of a Node pointer.

TBH, I'm not even sure what would be recommended as best practice. I agree it would be good to write down some recommendations (perhaps in rclcpp itself).

audrow:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14662/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9415/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12326/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14806/)

PR #57, opened 2021-06-17
Add missing runtime dependencies in package.xml

hidmic:
Precisely what the title says.

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -18,9 +18,9 @@
   <buildtool_depend>ament_cmake_python</buildtool_depend>
   <buildtool_depend>python_cmake_module</buildtool_depend>
 
-  <build_depend>builtin_interfaces</build_depend>
-  <build_depend>rclcpp</build_depend>
-  <build_depend>rclpy</build_depend>
+  <depend>builtin_interfaces</depend>
+  <depend>rclcpp</depend>
+  <depend>rclpy</depend>
 
   <test_depend>ament_lint_auto</test_depend>
   <test_depend>ament_cmake_gtest</test_depend>


package.xml
```diff
@@ -18,9 +18,9 @@
   <buildtool_depend>ament_cmake_python</buildtool_depend>
   <buildtool_depend>python_cmake_module</buildtool_depend>
 
-  <build_depend>builtin_interfaces</build_depend>
-  <build_depend>rclcpp</build_depend>
-  <build_depend>rclpy</build_depend>
+  <depend>builtin_interfaces</depend>
```
clalancette, line 21: As far as I can tell, `builtin_interfaces` is only ever used by the Python code in here.  Thus, I think it should be marked as an `exec_depend`, not a `depend`.

package.xml
```diff
@@ -18,9 +18,9 @@
...
+  <depend>rclcpp</depend>
```
clalancette, line 22: `rclcpp` isn't actually a `build_depend`, though it is required in the headers and in the tests.  So I think this should probably be:

```
<build_export_depend>rclcpp</build_export_depend>
<test_depend>rclcpp</test_depend>
```

package.xml
```diff
@@ -18,9 +18,9 @@
...
+  <depend>rclpy</depend>
```
clalancette, line 23: Similar to `builtin_interfaces`, `rclpy` is only ever used in the Python code, so it should be an `exec_depend`.

clalancette (changes requested):

package.xml
hidmic: You're absolutely right, which means the CMakeLists.txt is nonsense as well.

package.xml
hidmic: Agreed.

package.xml
hidmic: Indeed.

hidmic:
@clalancette I'll close this PR and open another, more comprehensive review of this package's dependencies.

PR #58, opened 2021-06-17
Correct package.xml and CMakeLists.txt

hidmic:
This patch revisits and corrects this package's metadata and build system. Namely,

- Use and export build and run time dependencies properly.
- Install Python code as a Python package.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,7 +15,6 @@ find_package(ament_cmake_python REQUIRED)
 find_package(ament_cmake_ros REQUIRED)
 find_package(rclcpp REQUIRED)
 find_package(rcutils REQUIRED)
-find_package(builtin_interfaces REQUIRED)
 
 add_library(${PROJECT_NAME} src/connection.cpp)
 if(WIN32)
@@ -29,8 +28,7 @@ target_include_directories(${PROJECT_NAME} PUBLIC
 ament_target_dependencies(${PROJECT_NAME}
   "rclcpp"
   "rcutils"
-  "builtin_interfaces"
-  )
+)
 
 install(
   TARGETS ${PROJECT_NAME} EXPORT ${PROJECT_NAME}
@@ -44,19 +42,13 @@ install(
   DESTINATION include
 )
 
-find_package(python_cmake_module REQUIRED)
-_ament_cmake_python_register_environment_hook()
-install(
-  DIRECTORY "src/${PROJECT_NAME}/"
-  DESTINATION "${PYTHON_INSTALL_DIR}/${PROJECT_NAME}"
-  PATTERN "*.pyc" EXCLUDE
-  PATTERN "__pycache__" EXCLUDE
-)
+ament_python_install_package(${PROJECT_NAME}
+  PACKAGE_DIR src/${PROJECT_NAME})
 
 ament_export_include_directories(include)
 ament_export_libraries(${PROJECT_NAME})
 ament_export_targets(${PROJECT_NAME})
-ament_export_dependencies(rclcpp rcutils builtin_interfaces)
+ament_export_dependencies(rclcpp rcutils)
 
 if(BUILD_TESTING)
   find_package(sensor_msgs REQUIRED)
@@ -120,6 +112,7 @@ if(BUILD_TESTING)
   endif()
 
   # Provides PYTHON_EXECUTABLE_DEBUG
+  find_package(python_cmake_module REQUIRED)
   find_package(PythonExtra REQUIRED)
   set(_PYTHON_EXECUTABLE "${PYTHON_EXECUTABLE}")
   if(WIN32)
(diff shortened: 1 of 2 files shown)


hidmic:
@clalancette It looks like `build_export_depend` tags are not picked by the `build and install` step in CI. Compare https://build.ros2.org/job/Rpr__message_filters__ubuntu_focal_amd64/18/console#console-section-3 (failing job) with https://build.ros2.org/job/Rpr__message_filters__ubuntu_focal_amd64/17/consoleFull#console-section-10 (passing job).

CMakeLists.txt
```diff
@@ -44,19 +42,13 @@ install(
...
-  PATTERN "__pycache__" EXCLUDE
-)
+ament_python_install_package(${PROJECT_NAME}
+  PACKAGE_DIR src/${PROJECT_NAME})
 
 ament_export_include_directories(include)
 ament_export_libraries(${PROJECT_NAME})
 ament_export_targets(${PROJECT_NAME})
-ament_export_dependencies(rclcpp rcutils builtin_interfaces)
+ament_export_dependencies(rclcpp rcutils)
```
clalancette, line 51: So this is a weird thing; what does `ament_export_dependencies` really mean?  Because technically this *package* depends on `builtin_interfaces`, but the C++ *library* does not.

clalancette (commented):
> @clalancette It looks like build_export_depend tags are not picked by the build and install step in CI. Compare https://build.ros2.org/job/Rpr__message_filters__ubuntu_focal_amd64/18/console#console-section-3 (failing job) with https://build.ros2.org/job/Rpr__message_filters__ubuntu_focal_amd64/17/consoleFull#console-section-10 (passing job).
>
> Is it intentional or is it a bug?

I *believe* it is intentional, but maybe @nuclearsandwich can provide more context.

Assuming it is intentional, I think the problem is that we shouldn't be doing the `find_package(rclcpp)` during the build phase, but only in the test phase.  That is, it isn't required for `rclcpp` to be there during build, but it is required during tests, and for any downstream packages that depend on it (hence the `test_depend` and the `build_export_depend`).  Does that make sense?

hidmic:
> Assuming it is intentional, I think the problem is that we shouldn't be doing the `find_package(rclcpp)` during the build phase, but only in the test phase. That is, it isn't required for `rclcpp` to be there during build, but it is required during tests, and for any downstream packages that depend on it (hence the `test_depend` and the `build_export_depend`). Does that make sense?

The thing is, we still have to invoke `ament_export_dependencies()` for `rclcpp` to be exported for downstream CMake package consumption. And that has to happen regardless of whether it was built for testing or not.

CMakeLists.txt
hidmic: `ament_export_dependencies` is limited to propagating dependencies for consumption within CMake. So we could argue that while `builtin_interfaces` is a dependency of `message_filters`, it is not a dependency of the CMake part of `message_filters`.

I'm trying to think of a case where that breaks, but I can't come up with any realistic one.

package.xml
```diff
@@ -14,17 +14,22 @@
...
+  <build_export_depend>rclcpp</build_export_depend>
+  <build_export_depend>rcutils</build_export_depend>
+
+  <exec_depend>builtin_interfaces</exec_depend>
+  <exec_depend>rclpy</exec_depend>
 
   <test_depend>ament_lint_auto</test_depend>
   <test_depend>ament_cmake_gtest</test_depend>
   <test_depend>ament_cmake_pytest</test_depend>
+  <test_depend>rclcpp</test_depend>
```
clalancette, line 32: One nit: now that it is a `build_depend`, I think you can drop the `test_depend`.

hidmic: Done in d7cc41a75b0748b42653e6e6a42f9692ba1852eb.

hidmic:
CI up to `message_filters` and `tf2_ros` (a dependent package):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14787/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9536/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12453/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14934/)

clalancette (approved):
The changes now seem good; thanks for going over it with me.

I'd say let's do a full `--packages-above-and-dependencies` CI build, just to make sure we get as much coverage as possible.

hidmic:
> I'd say let's do a full `--packages-above-and-dependencies` CI build, just to make sure we get as much coverage as possible.

Sounds reasonable:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14789/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9538/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12455/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14936/)

hidmic:
Alright, all green. Going in. Thanks @clalancette !

PR #67, opened 2021-11-19
Update maintainers

audrow:
Update maintainers to Geoffrey Biggs (@gbiggs).

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -6,13 +6,17 @@
   <description>
     A set of ROS2 message filters which take in messages and may output those messages at a later time, based on the conditions that filter needs met.
   </description>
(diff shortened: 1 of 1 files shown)


package.xml
```diff
@@ -6,13 +6,17 @@
   <description>
     A set of ROS2 message filters which take in messages and may output those messages at a later time, based on the conditions that filter needs met.
   </description>
-  <maintainer email="ethan.gao@linux.intel.com">Ethan Gao</maintainer>
```
clalancette, line 9: Since we haven't discussed this change directly with Ethan, let's not make this change for now.

clalancette (changes requested):

clalancette (approved):

PR #72, opened 2022-02-24
Use RCL_ROS_TIME for message_traits::TimeStamp 

### LINKED ISSUE #32: Message Cache Default Interpolation is RCL_SYSTEM_TIME
I was experimenting with using message filter caches to lookup message data and hit an odd behavior where the interpolation time always needs to be RCL_SYSTEM_TIME.   Otherwise you will get a:

`terminate called after throwing an instance of 'std::runtime_error'
  what():  can't compare times with different time sources`

It is my understanding that ROS2 design documents intend the preferred timestamp type to be RCL_ROS_TIME. Is this correct?

If so, I can put a pull request in for a one line change of:
https://github.com/ros2/message_filters/blob/a18a34a2242e6d44dd6c66e4cb2679d0fe07a40b/include/message_filters/message_traits.h#L89

to:

`return rclcpp::Time(stamp);`

This should use the message contructor instead of the sec/nanosec contructor, which defaults to RCL_ROS_TIME.

Alternately:

`return rclcpp::Time(stamp.sec, stamp.nanosec, RCL_ROS_TIME);`

ijnek:
When creating TimeStamp, make sure RCL_ROS_TIME is being used, and not RCL_SYSTEM_TIME.

Closes #32.

Signed-off-by: Kenji Brameld <kenjibrameld@gmail.com>

### DIFF
--- a/include/message_filters/message_traits.h
+++ b/include/message_filters/message_traits.h
@@ -85,7 +85,7 @@ template<typename M>
 struct TimeStamp<M, typename std::enable_if<HasHeader<M>::value>::type >
 {
   static rclcpp::Time value(const M& m) {
-    return rclcpp::Time(m.header.stamp);
+    return rclcpp::Time(m.header.stamp, RCL_ROS_TIME);
   }
 };
 
--- a/include/message_filters/message_traits.h
+++ b/include/message_filters/message_traits.h
@@ -85,8 +85,7 @@ template<typename M>
 struct TimeStamp<M, typename std::enable_if<HasHeader<M>::value>::type >
 {
   static rclcpp::Time value(const M& m) {
-    auto stamp = m.header.stamp;
-    return rclcpp::Time(stamp.sec, stamp.nanosec);
+    return rclcpp::Time(m.header.stamp);
   }
 };
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -115,6 +115,12 @@ if(BUILD_TESTING)
     ament_target_dependencies(${PROJECT_NAME}-test_fuzz "rclcpp" "sensor_msgs")
   endif()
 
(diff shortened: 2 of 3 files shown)


include/message_filters/message_traits.h
```diff
@@ -85,8 +85,7 @@ template<typename M>
 struct TimeStamp<M, typename std::enable_if<HasHeader<M>::value>::type >
 {
   static rclcpp::Time value(const M& m) {
-    auto stamp = m.header.stamp;
-    return rclcpp::Time(stamp.sec, stamp.nanosec);
+    return rclcpp::Time(m.header.stamp);
```
SivertHavso, line 88: Nitpicking, but if we not want to rely on the default constructor, we could use:
```suggestion
    return rclcpp::Time(m.header.stamp, RCL_ROS_TIME);
```
Also makes it clearer to understand what time is being used.
https://docs.ros2.org/galactic/api/rclcpp/classrclcpp_1_1Time.html#a13b8e8297e89516806b18c26747804d5

gbiggs:
@SivertHavso Can you please add a test that verifies the new behaviour?

ijnek:
@gbiggs
What's better, a test to check that it simply returns a RCL_ROS_TIME clock type, or a test to make sure it doesn't throw the exception (using `EXPECT_NO_THROW`)?

gbiggs:
Ideally both. That would ensure future changes don't unexpectedly break existing users.

ijnek:
@gbiggs Tests have been added.

gbiggs (approved):

gbiggs:
CI:

- Linux [Build Status](https://ci.ros2.org/job/ci_linux/16290/)
- Linux-aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/10910/)
- Windows [Build Status](https://ci.ros2.org/job/ci_windows/16657/)

PR #73, opened 2022-05-13
Add latest time zero-order-hold sync policy

andermi:
Rate logic added to #38.
Computes rates of received messages and pivots on fastest topic; essentially, upsamples slower topics with a zero-order hold.

### DIFF
--- a/include/message_filters/sync_policies/latest_time.h
+++ b/include/message_filters/sync_policies/latest_time.h
@@ -1,0 +1,278 @@
+/*********************************************************************
+* Software License Agreement (BSD License)
+*
+*  Copyright (c) 2009, Willow Garage, Inc.
+*  All rights reserved.
+*
+*  Redistribution and use in source and binary forms, with or without
+*  modification, are permitted provided that the following conditions
+*  are met:
+*
+*   * Redistributions of source code must retain the above copyright
+*     notice, this list of conditions and the following disclaimer.
+*   * Redistributions in binary form must reproduce the above
+*     copyright notice, this list of conditions and the following
+*     disclaimer in the documentation and/or other materials provided
+*     with the distribution.
+*   * Neither the name of the Willow Garage nor the names of its
+*     contributors may be used to endorse or promote products derived
+*     from this software without specific prior written permission.
+*
+*  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+*  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+*  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+*  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+*  COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+*  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+*  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+*  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+*  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+*  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+*  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+*  POSSIBILITY OF SUCH DAMAGE.
+*********************************************************************/
+
+#ifndef MESSAGE_FILTERS__SYNC_LATEST_TIME_H_
+#define MESSAGE_FILTERS__SYNC_LATEST_TIME_H_
+
+#include <cassert>
+#include <deque>
+#include <string>
+#include <tuple>
+#include <vector>
+#include <numeric>
+#include <algorithm>
+
+#include <inttypes.h>
+
+#include <rclcpp/rclcpp.hpp>
+#include <rcutils/logging_macros.h>
+
+#include "message_filters/connection.h"
+#include "message_filters/message_traits.h"
+#include "message_filters/null_types.h"
+#include "message_filters/signal9.h"
+#include "message_filters/synchronizer.h"
+
+
+#ifndef RCUTILS_ASSERT
+// TODO(tfoote) remove this after it's implemented upstream
+// https://github.com/ros2/rcutils/pull/112
+#define RCUTILS_ASSERT assert
+#endif
+#ifndef RCUTILS_BREAK
+#include <cassert>
+// TODO(tfoote) remove this after it's implemented upstream
+// https://github.com/ros2/rcutils/pull/112
+#define RCUTILS_BREAK std::abort
+#endif
+// Uncomment below intead
+//#include <rcutils/assert.h>
+
+namespace message_filters
+{
+namespace sync_policies
+{
+
+template<typename M0, typename M1,
+         typename M2 = NullType, typename M3 = NullType, typename M4 = NullType,
+         typename M5 = NullType, typename M6 = NullType, typename M7 = NullType,
+         typename M8 = NullType>
+struct LatestTime : public PolicyBase<M0, M1, M2, M3, M4, M5, M6, M7, M8>
+{
+  typedef Synchronizer<LatestTime> Sync;
+  typedef PolicyBase<M0, M1, M2, M3, M4, M5, M6, M7, M8> Super;
+  typedef typename Super::Messages Messages;
+  typedef typename Super::Signal Signal;
+  typedef typename Super::Events Events;
+  typedef typename Super::RealTypeCount RealTypeCount;
+  typedef typename Super::M0Event M0Event;
+  typedef typename Super::M1Event M1Event;
+  typedef typename Super::M2Event M2Event;
+  typedef typename Super::M3Event M3Event;
+  typedef typename Super::M4Event M4Event;
+  typedef typename Super::M5Event M5Event;
+  typedef typename Super::M6Event M6Event;
+  typedef typename Super::M7Event M7Event;
+  typedef typename Super::M8Event M8Event;
+  typedef std::deque<M0Event> M0Deque;
+  typedef std::deque<M1Event> M1Deque;
+  typedef std::deque<M2Event> M2Deque;
+  typedef std::deque<M3Event> M3Deque;
+  typedef std::deque<M4Event> M4Deque;
+  typedef std::deque<M5Event> M5Deque;
+  typedef std::deque<M6Event> M6Deque;
+  typedef std::deque<M7Event> M7Deque;
+  typedef std::deque<M8Event> M8Deque;
+  typedef std::vector<M0Event> M0Vector;
+  typedef std::vector<M1Event> M1Vector;
+  typedef std::vector<M2Event> M2Vector;
+  typedef std::vector<M3Event> M3Vector;
+  typedef std::vector<M4Event> M4Vector;
+  typedef std::vector<M5Event> M5Vector;
+  typedef std::vector<M6Event> M6Vector;
+  typedef std::vector<M7Event> M7Vector;
+  typedef std::vector<M8Event> M8Vector;
+  typedef Events Tuple;
+  typedef std::tuple<M0Deque, M1Deque, M2Deque, M3Deque, M4Deque,
+                     M5Deque, M6Deque, M7Deque, M8Deque> DequeTuple;
+  typedef std::tuple<M0Vector, M1Vector, M2Vector, M3Vector, M4Vector,
+                     M5Vector, M6Vector, M7Vector, M8Vector> VectorTuple;
+
+  LatestTime(uint32_t)
+  : parent_(0),
+    ros_clock_(RCL_ROS_TIME)  // RCL_STEADY_TIME)
+  {
+  }
+
+  LatestTime(const LatestTime& e)
+  {
+    *this = e;
+  }
+
+  LatestTime& operator=(const LatestTime& rhs)
+  {
+    parent_ = rhs.parent_;
+    events_ = rhs.events_;
+    rates_ = rhs.rates_;
+    ros_clock_ = rhs.ros_clock_;
+
+    return *this;
+  }
+
+  void initParent(Sync* parent)
+  {
+    parent_ = parent;
+  }
+
+  template<int i>
+  void add(const typename std::tuple_element<i, Events>::type& evt)
+  {
+    RCUTILS_ASSERT(parent_);
+    
+    std::lock_guard<std::mutex> lock(data_mutex_);
+    
+    if(!received_msg<i>())
+    {
+      rates_.push_back(Rate(1U, ros_clock_.now()));
+      // wait until we get each message once to publish
+      // then wait until we got each message twice to compute rates
+      // NOTE: this will drop a few messages of the faster topics until
+      //       we get one of the slowest so we can sync
+      std::get<i>(events_) = evt;  // adding here ensures we see even the slowest
+                                   // message twice before computing rate
+      return;
+    }
+
+    std::get<i>(events_) = evt;
+
+    compute_hz(rates_[i]);
+    if(i == find_pivot() && is_full())
+    {
+      publish();
+    }
+  }
+
+private:
+  // assumed data_mutex_ is locked
+  void publish()
+  {
+    parent_->signal(std::get<0>(events_), std::get<1>(events_), std::get<2>(events_),
+                    std::get<3>(events_), std::get<4>(events_), std::get<5>(events_),
+                    std::get<6>(events_), std::get<7>(events_), std::get<8>(events_));
+  }
+
+  struct Rate
+  {
+    uint64_t count;
+    rclcpp::Time start;
+    double hz;
+    Rate(const uint64_t &count, const rclcpp::Time &start)
+      : count(count),
+        start(start),
+        hz(0.0)
+    {
+    }
+    bool operator>(const Rate &that) const
+    {
+      return this->hz > that.hz;
+    }
+  };
+
+  // assumed data_mutex_ is locked
+  void compute_hz(Rate &rate)
+  {
+    rate.count += 1;
+    rclcpp::Time now = ros_clock_.now();
+    double elapsed = (now-rate.start).seconds();
+    rate.hz = double(rate.count)/elapsed;
+
+    if(rate.count % 10000U == 0U)
+    {
+      rate.count = 0U;
+      rate.start = ros_clock_.now();
+    }
+  }
+
+  template <typename T>
+  std::vector<std::size_t> sort_indices(const std::vector<T> &v)
+  {
+
+    // initialize original index locations
+    std::vector<std::size_t> idx(v.size());
+    std::iota(idx.begin(), idx.end(), 0U);
+
+    // sort indexes based on comparing values in v
+    // using std::stable_sort instead of std::sort
+    // to avoid unnecessary index re-orderings
+    // when v contains elements of equal values 
+    std::stable_sort(idx.begin(), idx.end(),
+                     [&v](std::size_t i1, std::size_t i2) {return v[i1] > v[i2];});
+
+    return idx;
+  }
+
+  // assumed data_mutex_ is locked
+  template<int i>
+  bool received_msg()
+  {
+    return (RealTypeCount::value > i ? (bool)std::get<i>(events_).getMessage() : true);
+  }
+
+  // assumed data_mutex_ is locked
+  bool is_full()
+  {
+    bool full = received_msg<0>();
+    full = full && received_msg<1>();
+    full = full && received_msg<2>();
+    full = full && received_msg<3>();
+    full = full && received_msg<4>();
+    full = full && received_msg<5>();
+    full = full && received_msg<6>();
+    full = full && received_msg<7>();
+    full = full && received_msg<8>();
+
+    return full;
+  }
+
+  // assumed data_mutex_ is locked
+  int find_pivot()
+  {
+    // find arg max rate
+    return sort_indices(rates_)[0U];
+  }
+
+  Sync* parent_;
+  static const int NO_PIVOT = 9;  // Special value for the pivot indicating
+                                  // that no pivot has been selected
+  Events events_;
+  std::vector<Rate> rates_;
+  std::mutex data_mutex_;  // Protects all of the above
+  
+  rclcpp::Clock ros_clock_;
+};
+
+}  // namespace sync
+}  // namespace message_filters
+
+#endif // MESSAGE_FILTERS__SYNC_LATEST_TIME_H_
--- a/include/message_filters/sync_policies/latest_time.h
+++ b/include/message_filters/sync_policies/latest_time.h
@@ -344,12 +344,17 @@ private:
           continue;
         }
       }
-      double rate_delta = rates_[pivot].hz - 1.0/period;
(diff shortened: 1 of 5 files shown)


clalancette (changes requested):
I've left a few comments inline.

Besides that, it might be nice to add in some documentation to the top of `latest_time.h` that describes briefly what this sync policy does, and how to use it.

include/message_filters/sync_policies/latest_time.h
```diff
@@ -0,0 +1,278 @@
...
+// TODO(tfoote) remove this after it's implemented upstream
+// https://github.com/ros2/rcutils/pull/112
+#define RCUTILS_ASSERT assert
+#endif
+#ifndef RCUTILS_BREAK
+#include <cassert>
+// TODO(tfoote) remove this after it's implemented upstream
+// https://github.com/ros2/rcutils/pull/112
+#define RCUTILS_BREAK std::abort
+#endif
```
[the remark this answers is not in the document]
andermi: removed

include/message_filters/sync_policies/latest_time.h
```diff
@@ -0,0 +1,278 @@
...
+#ifndef MESSAGE_FILTERS__SYNC_LATEST_TIME_H_
+#define MESSAGE_FILTERS__SYNC_LATEST_TIME_H_
+
+#include <cassert>
+#include <deque>
+#include <string>
+#include <tuple>
+#include <vector>
+#include <numeric>
+#include <algorithm>
```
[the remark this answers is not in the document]
andermi: alphabetized, and removed unused

test/test_latest_time_policy.cpp
```diff
@@ -0,0 +1,239 @@
...
+  }
+}
+
+int main(int argc, char **argv){
+  testing::InitGoogleTest(&argc, argv);
+  rclcpp::init(argc, argv);
+  return RUN_ALL_TESTS();
+}
+
+
```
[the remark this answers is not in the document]
andermi: removed

andermi:
Thanks @clalancette ! I added some some documentation at the top of the header.

test/test_latest_time_policy.cpp
```diff
@@ -0,0 +1,237 @@
...
+    {
+      sync.add<1>(q[idx/2]);
+    }
+    if(idx % 4U == 0U)
+    {
+      sync.add<2>(r[idx/4]);
+    }
+    sync.add<0>(p[idx]);
+    
+    ASSERT_EQ(h.count_, idx);
```
chapulina, line 132: I see that other tests in this repository make heavy use of asserts, but I'd recommend using `EXPECT`s in places where a failure won't cause the test to crash. This way when the test does fail, it still continues and can have other failures too. That can be useful to find out the root cause of the failure, because there's more information.

test/test_latest_time_policy.cpp
```diff
@@ -0,0 +1,237 @@
...
+    MsgPtr p_idx(std::make_shared<Msg>()); p_idx->data = idx; p.push_back(p_idx);
+    if(idx % 2U == 0U)
+    {
+      MsgPtr q_idx(std::make_shared<Msg>()); q_idx->data = idx; q.push_back(q_idx);
+    }
+    if(idx % 4U == 0U)
+    {
+      MsgPtr r_idx(std::make_shared<Msg>()); r_idx->data = idx; r.push_back(r_idx);
+    }
+  }
```
chapulina, line 171: Both tests are the same up to here. Consider moving the common code to a function. Maybe even to `LatestTime::SetUp` so it's called at the beginning of every test.

test/test_latest_time_policy.cpp
```diff
@@ -0,0 +1,237 @@
...
+    {
+      sync.add<1>(q[(idx-1U)/2U]);
+    }
+    if(idx % 4U == 3U)
+    {
+      sync.add<2>(r[(idx-3U)/4U]);
+    }
+    sync.add<0>(p[idx]);
+    
+    switch(idx)
```
chapulina, line 186: Consider condensing this entire switch block into something like this

```
    if (idx > 2U)
    {
      ASSERT_EQ(h.count_, idx - 2);
      ASSERT_EQ(h.p_->data, p[idx]->data);
      ASSERT_EQ(h.q_->data, q[(idx - 1) / 2]->data);
      ASSERT_EQ(h.r_->data, r[(idx - 3) / 4]->data);
    }
    else
    {
      ASSERT_FALSE(h.p_);
      ASSERT_FALSE(h.q_);
      ASSERT_FALSE(h.r_);
    }
```

include/message_filters/sync_policies/latest_time.h
```diff
@@ -0,0 +1,283 @@
...
+
+  // assumed data_mutex_ is locked
+  void compute_hz(Rate &rate)
+  {
+    rate.count += 1;
+    rclcpp::Time now = ros_clock_.now();
+    double elapsed = (now-rate.start).seconds();
+    rate.hz = double(rate.count)/elapsed;
+
+    if(rate.count % 10000U == 0U)
```
chapulina, line 215: I'm trying to understand the motivation for resetting every 10k calls. Maybe it would be smoother to use a moving window and always look back at the Nth previous call?

Also, maybe `compute_hz` could be a member of `Rate`, but you'd have to pass in the current time...

include/message_filters/sync_policies/latest_time.h
```diff
@@ -0,0 +1,283 @@
...
+  typedef std::vector<M4Event> M4Vector;
+  typedef std::vector<M5Event> M5Vector;
+  typedef std::vector<M6Event> M6Vector;
+  typedef std::vector<M7Event> M7Vector;
+  typedef std::vector<M8Event> M8Vector;
+  typedef Events Tuple;
+  typedef std::tuple<M0Deque, M1Deque, M2Deque, M3Deque, M4Deque,
+                     M5Deque, M6Deque, M7Deque, M8Deque> DequeTuple;
+  typedef std::tuple<M0Vector, M1Vector, M2Vector, M3Vector, M4Vector,
+                     M5Vector, M6Vector, M7Vector, M8Vector> VectorTuple;
```
chapulina, line 125: I don't see these tuples being used, are they needed?

include/message_filters/sync_policies/latest_time.h
```diff
@@ -0,0 +1,283 @@
...
+  // assumed data_mutex_ is locked
+  int find_pivot()
+  {
+    // find arg max rate
+    return sort_indices(rates_)[0U];
+  }
+
+  Sync* parent_;
+  static const int NO_PIVOT = 9;  // Special value for the pivot indicating
+                                  // that no pivot has been selected
```
chapulina, line 272: Is this supposed to be used somewhere?

chapulina (commented):
The functionality works for me. I have some questions and suggestions. Take it all with a grain of salt since it's my first time touching this repo 😄

test/test_latest_time_policy.cpp
andermi: changed to `EXPECT_*`

include/message_filters/sync_policies/latest_time.h
andermi: I changed to EMA on hz (removing `count`) with a heavy weight on new hz (0.9*new_hz + 0.1*old_hz) and it works, but I'm thinking this way I would I also need to detect a step change in message rates (in the event someone dynamically modifies pub rate) to reset the EMA.

include/message_filters/sync_policies/latest_time.h
andermi: removed

include/message_filters/sync_policies/latest_time.h
andermi: removed

test/test_latest_time_policy.cpp
andermi: I pushed it to a Test Fixture SetUp.

test/test_latest_time_policy.cpp
andermi: :+1:

include/message_filters/sync_policies/latest_time.h
andermi: I added another test to capture behavior of the policy and the EMA when pivot message changes rate and new pivot is selected.

Also, I moved `compute_h` to the `Rate` class.

test/test_latest_time_policy.cpp
```diff
@@ -0,0 +1,277 @@
...
+    {
+      EXPECT_EQ(h.count_, idx - 2U);
+      EXPECT_EQ(h.p_->data, p[idx]->data);
+      EXPECT_EQ(h.q_->data, q[(idx - 1U) / 2U]->data);
+      EXPECT_EQ(h.r_->data, r[(idx - 3U) / 4U]->data);
+    }
+    // rate of p still 50Hz @ idx==4, then change rate of p lower than q
+    // will not publish again until idx==9
+    // since rate of q is 25Hz, p was 50Hz @ idx==4,
+    // and new rate of p computed after q received @ idx==7
```
chapulina, line 236: I'm a bit confused about this test case. If I understand correctly, all 3 (p, q and r) have new values at `idx==7`. But that's not being published until `idx==9`. 

I understood that the filter was supposed to publish at the fastest rate possible, whenever there's a new message. So I'd expect it to publish at `idx==5` (when q changes), and again at `idx==7` (when they all change), so the subscriber always has the most up-to-date information.

Is my assumption correct?

andermi: I could try to change the behavior, but the last computed rate for `p` is 50Hz until it is received again when it is received at `idx==7` but it is received after `q` at `idx==7`. `q` would be the new pivot, but the policy doesn't know that yet until `q` is received at `idx==9`. I could change this to detect that `p` is late, but I'd have to think about side-effects and what to do if it's late.

andermi: > I'd expect it to publish at `idx==5` (when q changes)

The computed rate for `p` is still 50Hz at `idx==5` since it was computed as 50Hz at `idx==4` and won't be computed again until `idx==7`.

andermi: I now loop over the sorted rates and pick the first message that isn't late as the pivot. This has the effect of publishing on `idx==7`

andermi: Basically, q now detects that p is late at idx==7, but can't at idx==5 since p still has a chance to arrive after q and still be on time.

andermi: > publish at the fastest rate possible, whenever there's a new message

This would have the effect of publishing at a fairly erratic rate. For example, if `p` and `q` were both 50Hz, we'd get two publishes of all messages. If `p` and `q` came in 0.01s apart, everything would look like a ZOH @ 100Hz. If they came in closer/further apart, the publishing would be galloping. The publishing would be much more erratic for more messages.

chapulina: Thanks for the explanation, the consequences of publishing all the time are clear to me now 👍🏽

chapulina (approved):
Thanks for iterating, LGTM! I'd appreciate a review from one of the maintainers though.

include/message_filters/sync_policies/latest_time.h
```diff
@@ -0,0 +1,276 @@
...
+struct LatestTime : public PolicyBase<M0, M1, M2, M3, M4, M5, M6, M7, M8>
+{
+  typedef Synchronizer<LatestTime> Sync;
+  typedef PolicyBase<M0, M1, M2, M3, M4, M5, M6, M7, M8> Super;
+  typedef typename Super::Messages Messages;
+  typedef typename Super::Signal Signal;
+  typedef typename Super::Events Events;
+  typedef typename Super::RealTypeCount RealTypeCount;
+
+  LatestTime(uint32_t)
```
clalancette, line 95: What's this unused argument for?  Can we just remove it?

include/message_filters/sync_policies/latest_time.h
```diff
@@ -0,0 +1,276 @@
...
+      } else {
+        if (do_error_init) {
+          error = fabs(hz - 1.0/period);
+          do_error_init = false;
+        } else {
+          if (fabs(hz - 1.0/period) > 10.0*error) {
+            // detected step change in rate so reset
+            do_hz_init = true;
+            do_error_init = true;
+            compute_hz(now);
```
clalancette, line 191: I don't love the fact that this is a recursive function.  I guess in the nominal case it should only ever recurse once, but it makes it harder to reason about.  Could we possibly make this a loop instead?

include/message_filters/sync_policies/latest_time.h
```diff
@@ -0,0 +1,276 @@
...
+            return;
+          }
+          error = 0.7*error + 0.3*fabs(hz - 1.0/period);
+        }
+        hz = 0.9/period + 0.1*hz;
```
clalancette, line 196: Similarly, where do these magic numbers come from, and should they be configurable?

clalancette (changes requested):
Overall, looks pretty good to me.  I've left a few things inline that I think we should address.

include/message_filters/sync_policies/latest_time.h
andermi: yep! removed

include/message_filters/sync_policies/latest_time.h
andermi: turned it into a do-while

include/message_filters/sync_policies/latest_time.h
```diff
@@ -0,0 +1,276 @@
...
+      double period = (now-prev).seconds();
+      if (do_hz_init) {
+        hz = 1.0/period;
+        do_hz_init = false;
+      } else {
+        if (do_error_init) {
+          error = fabs(hz - 1.0/period);
+          do_error_init = false;
+        } else {
+          if (fabs(hz - 1.0/period) > 10.0*error) {
```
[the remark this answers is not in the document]
andermi: config now

include/message_filters/sync_policies/latest_time.h
andermi: also configs

andermi:
@clalancette I think/hope I've addressed your comments.

include/message_filters/sync_policies/latest_time.h
```diff
@@ -0,0 +1,375 @@
...
+      //       we get one of the slowest so we can sync
+      std::get<i>(events_) = evt;  // adding here ensures we see even the slowest
+                                   // message twice before computing rate
+      return;
+    }
+
+    std::get<i>(events_) = evt;
+    rclcpp::Time now = ros_clock_->now();
+    std::cout << "compute rate for [" << i << "]" << std::endl;
+    rates_[i].compute_hz(now);
```
clalancette, line 172: Please remove this debugging statement.

include/message_filters/sync_policies/latest_time.h
```diff
@@ -0,0 +1,375 @@
...
+    if(i == find_pivot(now) && is_full())
+    {
+      std::cout << "I'm pivot! [" << i << "] rate [" << rates_[i].hz << "]" << std::endl;
```
clalancette, line 175: Ditto.

include/message_filters/sync_policies/latest_time.h
```diff
@@ -0,0 +1,375 @@
...
+      publish();
+    } else {
+      std::cout << "I'm not pivot... [" << i << "] rate [" << rates_[i].hz << "]" << std::endl;
+    }
```
clalancette, line 179: Ditto.

include/message_filters/sync_policies/latest_time.h
```diff
@@ -0,0 +1,375 @@
...
+      const double & rate_ema_alpha, const double & error_ema_alpha,
+      const double & rate_step_change_margin_factor)
+      : prev{start},
+        rate_ema_alpha{rate_ema_alpha},
+        error_ema_alpha{error_ema_alpha},
+        rate_step_change_margin_factor{rate_step_change_margin_factor}
+    {
+      std::cout << "prev{" << prev.seconds() << 
+        "}, rate_ema_alpha{" << rate_ema_alpha << "}, error_ema_alpha{" << error_ema_alpha <<
+        "}, rate_step_change_margin_factor{" << rate_step_change_margin_factor << "}" << std::endl;
```
clalancette, line 251: Ditto.

include/message_filters/sync_policies/latest_time.h
```diff
@@ -0,0 +1,375 @@
...
+    {
+      return this->hz > that.hz;
+    }
+
+    void compute_hz(const rclcpp::Time & now)
+    {
+      bool step_change_detected = false;
+      do {
+        double period = (now-prev).seconds();
+        RCUTILS_ASSERT(period > 0.0 && "Multiple messages received and time is not updating.");
```
clalancette, line 264: I'm not a huge fan of this.  This resolves into an `assert` statement, which means that when building in release mode, this protection won't be in place.  I don't think this is a very performance-sensitive piece of code, so I'd rather that we just make this an if statement that exits the function if we see it.

include/message_filters/sync_policies/latest_time.h
```diff
@@ -0,0 +1,375 @@
...
+
+  // assumed data_mutex_ is locked
+  int find_pivot(const rclcpp::Time & now)
+  {
+    // find arg max rate
+    std::vector<std::size_t> sorted_idx = sort_indices(rates_);
+
+    // use fastest message that isn't late as pivot
+    for (size_t pivot : sorted_idx) {
+      std::cout << "checking pivot [" << pivot << "]" << std::endl;
```
clalancette, line 341: Please remove this debugging statement.

include/message_filters/sync_policies/latest_time.h
```diff
@@ -0,0 +1,375 @@
...
+      double period = (now-rates_[pivot].prev).seconds();
+      if (period == 0.0) {
+        // we just updated updated this one,
+        // and it's fastest, so use as pivot
+        std::cout << "I just updated and I'm fastest" << std::endl;
```
clalancette, line 346: Please remove this debugging statement.

include/message_filters/sync_policies/latest_time.h
```diff
@@ -0,0 +1,375 @@
...
+#define MESSAGE_FILTERS__SYNC_LATEST_TIME_H_
+
+#include <algorithm>
+#include <cassert>
+#include <deque>
+#include <memory>
+#include <numeric>
+#include <string>
+#include <tuple>
+#include <vector>
```
clalancette, line 77: Given what is actually used in this header file, I think this whole block can be reduced to:

```
#include <algorithm>
#include <mutex>
#include <numeric>
#include <tuple>
#include <vector>
```

test/test_latest_time_policy.cpp
```diff
@@ -0,0 +1,344 @@
...
+*  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+*  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+*  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+*  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+*  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+*  POSSIBILITY OF SUCH DAMAGE.
+*********************************************************************/
+
+#include <gtest/gtest.h>
+
```
clalancette, line 36: Add:

```
#include <functional>
#include <memory>
#include <vector>
```

clalancette (changes requested):
All right, one last pass.  Once these things are fixed, we'll be ready to run CI on it.

include/message_filters/sync_policies/latest_time.h
andermi: sorry about all the debugs :disappointed: 
removed!

include/message_filters/sync_policies/latest_time.h
andermi: done

include/message_filters/sync_policies/latest_time.h
andermi: done

test/test_latest_time_policy.cpp
andermi: done

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16803/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11382/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17208/)

andermi:
I actually did find a small intermittent bug

clalancette:
Let's run Windows CI again to see where we are at: * Windows [Build Status](http://ci.ros2.org/job/ci_windows/17217/)

test/test_latest_time_policy.cpp
```diff
@@ -0,0 +1,331 @@
...
+      EXPECT_EQ(h.r_->data, r[(idx - 5U) / 4U]->data);
+    }
+    else
+    {
+      EXPECT_FALSE(h.p_);
+      EXPECT_FALSE(h.q_);
+      EXPECT_FALSE(h.r_);
+    }
+
+    rate.sleep();
```
chapulina, line 322: Instead of just sleeping a fixed amount of time between loops, is there a condition that could be checked here so it waits until that's satisfied? That would remove the reliance on time and make the test robust on slower machines.

One common pattern I've used is something like this:

```cpp
int sleep{0};
int maxSleep{30};
for (; condition && sleep < maxSleep; ++sleep)
{
  // sleep
}
EXPECT_LT(sleep, maxSleep);
EXPECT_TRUE(condition);
```

chapulina (commented):
It looks like there are some Windows warnings being introduced by this PR:

https://ci.ros2.org/job/ci_windows/17217/msbuild/new/#issuesContent

test/test_latest_time_policy.cpp
andermi: Well, the rate sleep is for creating message rates.

I suppose instead of using a rate sleep, I could pass the `Policy` an `rclcpp::Clock` that is set for sim time, and control the `/clock` in the test? Then the time the messages are received would be deterministic.

Requiring stamped messages could also resolve timing issues. I was hoping to make it generic, but users should really be using stamps for messages that require timing anyways :stuck_out_tongue_winking_eye:

quarkytale:
Retriggered CI
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16963/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11527/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17362/)

quarkytale:
This is currently blocked by timing issue in windows CI.
There are two EMA filters tuned for Linux: 
1. mean rate for incoming message
2. mean error from 1 which is similar to stddev 

In Linux, a threshold of 10 * EMA#2 is used for detecting if the message rate changed. On windows, the rate change is not being detected the same way.

andermi:
@clalancette I gave it another shot... I spun up ros2 dev on my windows machine.
The test passes using sim /clock to control the test timing

quarkytale:
Retriggered Windows CI [Build Status](https://ci.ros2.org/job/ci_windows/17557/)

andermi:
Looks like compiler warnings causing `unstable`. Working through them now.

andermi:
@quarkytale would you mind running another one? I fixed the 7 msbuild warnings related to conversion from `size_t`. Can you see anything else to fix first?

quarkytale:
I'm not sure if the warnings were an issue, but good to have them cleared. Seems like some tests are being skipped, that might be the reason for unstable status. Other things that stood out to me:
- `[WARNING] [python.exe-3]: 'SIGINT' sent to process[python.exe-3] not supported on Windows, escalating to 'SIGTERM'`
- `[-ERROR-] IO exception has been thrown: java.nio.file.NoSuchFileException: C:\ci\ws\src\ros2\message_filters\include\message_filters\sync_policies\latest_time.h`

Let's see if any of these persist: [Build Status](https://ci.ros2.org/job/ci_windows/17558/)

quarkytale (approved):
LGTM and CI is green!

adityapande-1995:
https://github.com/Mergifyio backport humble

Issue #75, opened 2022-06-15
Humble / Rolling in 22.04 Cannot handle callback groups


SteveMacenski:
We've been having an issue in Nav2 as of moving to 22.04 that our obstacle layers / costmap collision checkers haven't been working properly https://github.com/ros-planning/navigation2/issues/3014 -- e.g. no data is getting to our callbacks, which is quite the problem for a vision-based navigation system :laughing: 

It seems to boil down to the fact that when we moved to using a single set of nodes for each process, we starting making a bunch of use of callback groups that were tested and working fine on rolling in 20.04. I checked and Galactic is on 3.X.X of this package so rolling was probably around there when it was on 20.04, but now rolling is on 4.X.X along with Humble, which includes the changes in the PR below.

What appears to be the issue is that https://github.com/ros2/message_filters/pull/56 incorrectly added an error in one of the constructors for processing subscriber options which includes the callback group request. 

```
  virtual void subscribe(
    NodePtr node,
    const std::string& topic,
    const rmw_qos_profile_t qos,
    rclcpp::SubscriptionOptions options)
  {
    this->subscribe(node.get(), topic, qos, options);
  };

  virtual void subscribe(
    NodeType * node,
    const std::string& topic,
    const rmw_qos_profile_t qos,
    rclcpp::SubscriptionOptions options)
  {
    (void) options;
    RCLCPP_ERROR(
      node->get_logger(),
      "SubscriberBase::subscribe with four arguments has not been overridden");
    this->subscribe(node, topic, qos);
  }
``` 

So if the constructor is called with a node type pointer rather than a shared pointer then the `options` aren't passed and has a warning which seems bogus (?) since the constructor just above it does this fine with another pointer type. This entirely explains the behavior we're seeing if its using the second constructor and throwing out our options because of the type of the node provided. 

I believe the fix is to remove the warning and use the `option` in the `this->subscribe()` call in the second implementation.

PR #76, opened 2022-06-15
Adding fix to subscribe() call with raw node pointer and subscriber options

SteveMacenski:
Per https://github.com/ros2/message_filters/issues/75

### DIFF
--- a/include/message_filters/subscriber.h
+++ b/include/message_filters/subscriber.h
@@ -109,11 +109,7 @@ public:
     const rmw_qos_profile_t qos,
     rclcpp::SubscriptionOptions options)
   {
-    (void) options;
-    RCLCPP_ERROR(
-      node->get_logger(),
-      "SubscriberBase::subscribe with four arguments has not been overridden");
-    this->subscribe(node, topic, qos);
+    this->subscribe(node, topic, qos, options);
   }
 
   /**


gbiggs (approved):

SteveMacenski:
I'm 3 hours into the ~4 hours needed to build all of ROS 2 / underlay / nav2 from source to test this with the forked version of `message_filters` to make sure this is the only / correct issue. Hopefully I should be able to confirm in the next couple of hours, but either way this is an error.

SteveMacenski:
Well I've given up on this, there's a bunch of python conflicts with my installs  (launch won't launch, cli won't cli) and I'm not going to fix all of those issues for just testing this. https://github.com/ros2/rclcpp/issues/1954 identifies that there is also an issue in or below `rclcpp` but this is also a component of it.

SteveMacenski:
@gbiggs can we get this into a new Rolling / Humble release? That would help fix Nav2 in the next sync

gbiggs:
Backport to Humble and releases for Humble and Rolling are done! The fix should be in the next sync of each. Ping me if it doesn't appear.

ros-discourse:
This pull request has been mentioned on **ROS Discourse**. There might be relevant details there:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2022-06-16/26093/1

PR #79, opened 2022-07-14
Fix python examples and add a new example in documentation

charlielito:
Implemented suggestions of #50

### DIFF
--- a/index.rst
+++ b/index.rst
@@ -68,7 +68,8 @@ Another example for syncronizing one image topic and one pointcloud2 topic using
             # do your stuff here
             pass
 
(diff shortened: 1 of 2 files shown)


index.rst
```diff
@@ -36,10 +36,40 @@ Using the time synchronizer::
...
+                [self.image_sub, self.plc_sub],
+                queue_size,
+                0.01,  # defines the delay (in seconds) with which messages can be synchronized
+            )
+
+        def callback(self, image_msg, pcl_msg):
+            # do your stuff here
+            pass
+
+**Note**: It is **VERY IMPORTANT** that each subscriber has the same ``qos_profile`` than the one specified in the corresponding publisher code for each topic you want to subscribe to. If they don't match, the callback won't be executed (without any warning) and you will be very frustrated.
```
gbiggs, line 71:
```suggestion
**Note**: It is **VERY IMPORTANT** that each subscriber has the same ``qos_profile`` as the one specified in the corresponding publisher code for each topic you want to subscribe to.
If they don't match, the callback won't be executed (without any warning) and you will be very frustrated.
```

PR #40, opened 2019-10-14
Resolve ambiguity with boost/bind.hpp

sloretz:
This resolves an issue where `std::bind` placeholders are ambiguous when including both `boost/bind.hpp` and `message_filters/signal9.h`. `boost/bind.hpp` includes the line `using namespace boost::placeholders;`. Message filters also uses `using namespace std::placeholders` (in a smaller scope). That causes a compile error about the unqualified placeholders (`_1`, `_2`, ...) in `signal9.h` being ambiguous. This PR works around it by using the fully qualified namespace.

My motivation is porting `pcl_ros` to ROS 2. `pcl` 1.8 includes `boost/bind.hpp` in public headers, and `pcl_ros` uses `message_filters`.

<details><summary>pcl 1.8 files using boost bind</summary>

```
/usr/include/pcl-1.8/pcl/registration/boost.h:#include <boost/bind.hpp>
/usr/include/pcl-1.8/pcl/visualization/boost.h:#include <boost/bind.hpp>
/usr/include/pcl-1.8/pcl/filters/boost.h:#include <boost/bind.hpp>
/usr/include/pcl-1.8/pcl/keypoints/keypoint.h:#include <boost/bind.hpp>
/usr/include/pcl-1.8/pcl/io/boost.h:#include <boost/bind.hpp>
/usr/include/pcl-1.8/pcl/io/openni2/openni2_device.h:#include <boost/bind.hpp>
/usr/include/pcl-1.8/pcl/features/feature.h:#include <boost/bind.hpp>
/usr/include/pcl-1.8/pcl/features/boost.h:#include <boost/bind.hpp>
/usr/include/pcl-1.8/pcl/surface/boost.h:#include <boost/bind.hpp>
```
</details>

<details><summary>Compile errors in pcl_ros</summary>

```
/opt/ros/dashing/include/message_filters/signal9.h:281:126: error: reference to ‘_6’ is ambiguous
     return addCallback(std::function<void(P0, P1, P2, P3, P4, P5, P6, P7, NullP)>(std::bind(callback, t, _1, _2, _3, _4, _5, _6, _7, _8)));
                                                                                                                              ^~
In file included from /usr/include/boost/function.hpp:13:0,
                 from /usr/include/pcl-1.8/pcl/features/feature.h:48,
                 from /usr/include/pcl-1.8/pcl/features/normal_3d.h:44,
[...]

Issue #22, opened 2019-04-29
Signal9.h "Reference "_1" is ambigious"


klintan:
I use Boost in the parent project, including the message filters and get the "Reference "_1" is ambigious" everywhere where the placeholders is defined. Even though the `using namespace std::placeholders;` is defined in each function scope. 

 ```
template<typename P0, typename P1>
  Connection addCallback(void(*callback)(P0, P1))
  {
    using namespace std::placeholders;
    return addCallback(std::function<void(P0, P1, NullP, NullP, NullP, NullP, NullP, NullP, NullP)>(std::bind(callback, _1, _2)));
  }

  template<typename P0, typename P1, typename P2>
  Connection addCallback(void(*callback)(P0, P1, P2))
  {
    using namespace std::placeholders;
    return addCallback(std::function<void(P0, P1, P2, NullP, NullP, NullP, NullP, NullP, NullP)>(std::bind(callback, _1, _2, _3)));
  }
```

It should lookup the variable in the local namespace first I presume? as stated here https://stackoverflow.com/a/9381283/1829511 but it seems that the global namespace somehow takes precedence? Could this happen if somewhere any of my libs has a `using namespace boost` for instance? 
Any other ideas why this happens ? 

My solution was to define the fully qualified name  
```
std::placeholders::_1
``` 
but that is a bit of a pain.

dirk-thomas:
Are you invoking `using namespace`...` anywhere globally in a header in your code? If yes, than that is a bad idea and you might want to fix this in your code.

klintan:
I'll have an extra look in all headers and see if I can find any.

PR #87, opened 2022-12-11
Fix cache.h std::placeholder namespace

HaoruXue:
Same as in #22 which was resolved by #40, boost users would have problems due to ambiguous namespace issue with `message_filters::Cache`. The same change as in #40 should be applied to it.


clalancette (approved):

ahcorde:
https://github.com/Mergifyio backport humble

PR #93, opened 2023-04-09
Migrate to C++11 variadic templates

cwecht:
This PR mirgates message_synchronisers to variadic templates and allows for arbitrary numbers of topics.

Please note, that TimeSynchronizer had to be shruken down significantly for two reasons:

1. Implementing add0 .. add8 is now quiet hard. It has been dropped, because there is a generic alternative with add<I>
2. Variadic parameters are required to be at the end of a function signature. Therefore the queue_size parameter had to be moved to the front.

Given these changes, it might make sense to make TimeSynchronizer a template alias rather than a template class.

### DIFF
--- a/include/message_filters/time_synchronizer.hpp
+++ b/include/message_filters/time_synchronizer.hpp
@@ -73,23 +73,12 @@ void callback(const sensor_msgs::msg::CameraInfo::SharedPtr, const sensor_msgs::
 \endverbatim
  *
  */
-template<class M0, class M1, class M2 = NullType, class M3 = NullType, class M4 = NullType,
-  class M5 = NullType, class M6 = NullType, class M7 = NullType, class M8 = NullType>
-class TimeSynchronizer : public Synchronizer<sync_policies::ExactTime<M0, M1, M2, M3, M4, M5, M6,
-    M7, M8>>
+template<class ... Ms>
+class TimeSynchronizer : public Synchronizer<sync_policies::ExactTime<Ms...>>
 {
 public:
-  typedef sync_policies::ExactTime<M0, M1, M2, M3, M4, M5, M6, M7, M8> Policy;
+  typedef sync_policies::ExactTime<Ms...> Policy;
   typedef Synchronizer<Policy> Base;
-  typedef std::shared_ptr<M0 const> M0ConstPtr;
-  typedef std::shared_ptr<M1 const> M1ConstPtr;
-  typedef std::shared_ptr<M2 const> M2ConstPtr;
-  typedef std::shared_ptr<M3 const> M3ConstPtr;
-  typedef std::shared_ptr<M4 const> M4ConstPtr;
-  typedef std::shared_ptr<M5 const> M5ConstPtr;
-  typedef std::shared_ptr<M6 const> M6ConstPtr;
-  typedef std::shared_ptr<M7 const> M7ConstPtr;
-  typedef std::shared_ptr<M8 const> M8ConstPtr;
 
   using Base::add;
   using Base::connectInput;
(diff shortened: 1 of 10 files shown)


roncapat:
@mjcarroll any update on this?

destogl (approved):
This would be very nice to simplify and shorten the code!!

ahcorde (changes requested):
There are many conflicts

gavanderhoorn:
> There are many conflicts

Sort-of to be expected after 1+ years in the queue, right?

cwecht:
I'm happy to update this PR. I'll have a look at it soon.

ahcorde (changes requested):
cpplint is failing and also some tests

ahcorde (changes requested):
`depth_image_proc` in `image_pipeline` is not compiling

```cpp
...
  image_transport::SubscriberFilter sub_depth_image_;
  message_filters::Subscriber<sensor_msgs::msg::CameraInfo> sub_info_;
  using Sync = message_filters::TimeSynchronizer<sensor_msgs::msg::Image,
      sensor_msgs::msg::CameraInfo>;
  std::shared_ptr<Sync> sync_;
...

// this line is failing
sync_ = std::make_shared<Sync>(queue_size, sub_depth_image_, sub_info_);
```

Do you mind to take a look ?

ahcorde:
I made these changes https://github.com/ros-perception/image_pipeline/pull/1015 and it's working, I would perfer not to do it, to avoid affect other users.

cwecht:
I now reintroduced the TimeSynchronizers old constructors but with a deprecation marking. That's why their are new warnings in the build.

ahcorde (changes requested):
There are some new  warnings https://build.ros2.org/job/Rpr__message_filters__ubuntu_noble_amd64/83/clang-tidy/new/

cwecht:
> There are some new warnings https://build.ros2.org/job/Rpr__message_filters__ubuntu_noble_amd64/83/clang-tidy/new/

Yes, that's entended. The idea is to keep the old constructors in order to not break builds but mark them as "deprecated" as suggested in the [ROS 2 developer guide](https://docs.ros.org/en/rolling/The-ROS2-Project/Contributing/Developer-Guide.html#deprecation-strategy). Marking these constructors as deprecated causes the new warnings.

However, of cause we could just keep the old constructors and not mark them deprecated, but I suggest that we should remove them as they are quite redundant.

include/message_filters/time_synchronizer.hpp
```diff
@@ -73,156 +73,73 @@ void callback(const sensor_msgs::msg::CameraInfo::SharedPtr, const sensor_msgs::
...
     F0 & f0, F1 & f1, F2 & f2, F3 & f3, F4 & f4, F5 & f5, F6 & f6, F7 & f7,
     uint32_t queue_size)
-  : Base(Policy(queue_size))
-  {
-    connectInput(f0, f1, f2, f3, f4, f5, f6, f7);
-  }
+  : TimeSynchronizer(queue_size, f0, f1, f2, f3, f4, f5, f6, f7) {}
 
   template<class F0, class F1, class F2, class F3, class F4, class F5, class F6, class F7, class F8>
+  [[deprectead("Use variade constructor instead")]]
```
ahcorde, line 137:
```suggestion
  [[deprecated("Use variade constructor instead")]]
```

ahcorde (changes requested):
Totally fine with the deprecation, but the warning is related with the wronly written word `deprectead` -> `deprected`.

Anyhow If there are some warnings in the tests (potencially) you should use something like:

```cpp

// suppress deprecated warnings
#if !defined(_WIN32)
# pragma GCC diagnostic push
# pragma GCC diagnostic ignored "-Wdeprecated-declarations"
#else  // !defined(_WIN32)
# pragma warning(push)
# pragma warning(disable: 4996)
#endif

   // any calls to deprecated code

// remove warning suppression
#if !defined(_WIN32)
# pragma GCC diagnostic pop
#else  // !defined(_WIN32)
# pragma warning(pop)
#endif
```

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21576/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15865/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1074/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22296/)

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21592/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15879/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1093/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22310/)

ahcorde (approved):

ahcorde:
Thank you for the refactoring and the iterations

PR #94, opened 2023-05-03
feat: add signal time functions to ExactTime policy

russkel:
There was no easy way to obtain the last 'signal' time of the sync policy in order to determine the last successful sync time. This exposes the underlying time objects.

### DIFF
--- a/include/message_filters/sync_policies/exact_time.h
+++ b/include/message_filters/sync_policies/exact_time.h
@@ -138,6 +138,11 @@ struct ExactTime : public PolicyBase<M0, M1, M2, M3, M4, M5, M6, M7, M8>
     return drop_signal_.addCallback(callback, t);
   }
 
+  rclcpp::Time getLastSignalTime() const
+  {
+    return last_signal_time_;
+  }
+
 private:
 
   // assumes mutex_ is already locked


include/message_filters/sync_policies/exact_time.h
```diff
@@ -138,6 +138,16 @@ struct ExactTime : public PolicyBase<M0, M1, M2, M3, M4, M5, M6, M7, M8>
...
 
+  rclcpp::Time getLastSignalTime() const
+  {
+    return last_signal_time_;
+  }
+
+  rclcpp::Duration getElapsedSignalTime() const
+  {
+    return rclcpp::Clock().now() - last_signal_time_;
+  }
```
clalancette, line 149: This call is using a default clock, but that may or may not be the correct clock type that `last_signal_time` was collected from.  Additionally, taking the time at `now` doesn't give you the elapsed signal time; it gives you the time from when the last signal happened to right now.

Overall, I'd say remove this method.  You can always calculate this locally if you really want to given the `getLastSignalTime` above.

clalancette (changes requested):

include/message_filters/sync_policies/exact_time.h
russkel: > This call is using a default clock,

Right.

> Overall, I'd say remove this method. You can always calculate this locally if you really want to given the `getLastSignalTime` above.

When you say "if you really want to" - is there a better way of determining the last sync time? We use it as a "safety" check that things are working as expected.

clalancette: > When you say "if you really want to" - is there a better way of determining the last sync time? We use it as a "safety" check that things are working as expected.

Well, once you have `getLastSignalTime`, you can make this determination in your code.  There, you should have more context about which clock is in use.

russkel: Lovely. I will fix up the PR.

clalancette (approved):

PR #99, opened 2023-06-22
Fix python examples

ricardodeazambuja:
Nothing fancy, I just fixed the python examples where the order of the arguments "topic name"  and "message type" was inverted. I also added “round bracket” or “parentheses” to the print function.


clalancette (approved):
This looks good, thank you!

clalancette:
Merging without CI since this is updating documentation.

PR #101, opened 2023-08-03
Update the HasHeader check to be more specific.

clalancette:
That is, make sure that not only does the message have a field called 'header', but that it also is of type std_msgs/msg/Header.  Otherwise we won't be able to compile the bits below.


mjcarroll (approved):

mjcarroll:
LGTM with CI

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19335/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13863/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20056/)

Issue #31, opened 2019-07-01
std::unique_ptr not supported


AndreasAZiegler:
In order to get zero-copying inter-communication, ```std::unique_ptr``` are required. Unfortunately message_filters does not yet support ```std::unique_ptr```.

AndreasAZiegler:
Is there any update?

ZhenshengLee:
duplicate with https://github.com/ros2/message_filters/issues/28 which is already answered by https://answers.ros.org/question/325146/how-can-i-implement-efficient-intra-process-communication-with-message-filters/

ZhenshengLee:
can be closed.

tonynajjar:
I'm also finding myself needing this. Any update?

> can be closed.

I wouldn't close this issue, the answer provided only proposes a workaround, it would be nice to have a PR for this if it's a viable solution approved by the maintainers

roncapat:
I agree. I will try to implement if I have time, but I'm not making any real commitment, so feel free to do the same and open a PR pointing to this issue.

roncapat:
Moreover, be aware that rolling supports Type Adaptation (did a PR some time ago). This means that, in conjunction with intra-process comms, we would be able then to leverage the full spectrum of solutions to optimize wide ROS data exchange.

roncapat:
I'm giving it a look.
In my experience, I always used `const MessageType::ConstSharedPtr` as subscription callback type, and built `MessageType::UniquePtr` to be later `->publish(std::move(...))`d.
`message_filters` make use of `std::shared_ptr<MessageType const>` signature, which is equivalent to `MessageType::ConstSharedPtr`. 

It made be worth a try just to add `const` and see what happens as a first iteration.

roncapat:
To be completely honest, it seems to me that:
https://github.com/ros2/rclcpp/blob/9019a8d9b7710bd405a5a31c8d74fdc9ffa686dd/rclcpp/include/rclcpp/any_subscription_callback.hpp#L935
just requires the pointed object of the shared_ptr to be const, not necessarily the pointer itself. 

May I ask to you all to verify?

roncapat:
@tonynajjar @AndreasAZiegler may I ask you to test #103 and see if you get any performance gain?

tonynajjar:
Hey, I can try tomorrow. What should I be on the lookout for exactly? CPU consumption? Anything else?
As far as I know the msg memory address should be the same regardless of the change in the PR right?

[...]

PR #103, opened 2023-10-05
Mark subscription cb parameter const

roncapat:
See #31.
We can use `const ConstSharedPtr` here.

### DIFF
--- a/include/message_filters/subscriber.h
+++ b/include/message_filters/subscriber.h
@@ -307,7 +307,7 @@ public:
       qos_ = qos;
       options_ = options;
       sub_ = node->template create_subscription<M>(topic, rclcpp_qos,
-          [this](std::shared_ptr<const MessageType const> msg) {
+          [this](const std::shared_ptr<MessageType const> msg) {
             this->cb(EventType(msg));
           }, options);
 
--- a/include/message_filters/subscriber.h
+++ b/include/message_filters/subscriber.h
@@ -307,7 +307,7 @@ public:
       qos_ = qos;
       options_ = options;
       sub_ = node->template create_subscription<M>(topic, rclcpp_qos,
-          [this](std::shared_ptr<MessageType const> msg) {
+          [this](std::shared_ptr<const MessageType const> msg) {
             this->cb(EventType(msg));
           }, options);
 


ZhenshengLee (approved):
lg2m

clalancette:
Would it be possible to add a test showing this in use?  That is, add a test that would have failed before this PR, but now succeeds.

roncapat:
As pointed out in https://github.com/ros2/message_filters/issues/31#issuecomment-1749785441, it may not be beneficial in terms of performances.
Nevertheless, the `EventType` contructor takes a `const shared_ptr<>`:
https://github.com/ros2/message_filters/blob/856488d894657a296b249ebbc2feebeafcba8917/include/message_filters/message_event.h#L118C7-L118C7
so why not mark it `const` here, as the use of the lambda parameter gets only read, to remark this fact?

In the meantime, I'm trying to check with @tonynajjar if this also helps enabling Intra-process comms or not, because per this comment https://github.com/ros2/message_filters/issues/31#issuecomment-1749785441 I remembered that the possible signatures of intra-process compatible subscription callbacks were quite few, and in my code I took the habit to use `const MsgType::ConstSharedPtr` but looking into `rclcpp` now there seems to me that just using `MsgType::ConstSharedPtr` would still be fine.

In the latter case, `message_filters` is already IPC-enabled, unless I'm missing other bottlenecks / internal message copies somewhere else.

roncapat:
@clalancette I've changed title. Let me know if my last message rationale is good for you.

clalancette:
> In the latter case, `message_filters` is already IPC-enabled, unless I'm missing other bottlenecks / internal message copies somewhere else.

Yeah, I doubt this change is going to make much difference.  We are already using the `const ConstMessagePtr` signature of `MessageEvent` in any case, so passing it along like this is just a small hint to the compiler.  I don't think it is going to make a measurable performance difference.

That said, I'm not against adding it in.

roncapat:
As you want :) 

Yep, this is definitely to be considered as just a compiler hint (and developer hint, if someone reads the code), no additional feature.

clalancette:
I'll run CI on this, just to check that it is OK.  Assuming that is good, I'll merge it in:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19644/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14152/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20368/)

PR #104, opened 2023-11-13
Fixes pointed out by clang

clalancette:
1.  Make one of the SubscriptionBase::subscribe methods pure virtual.  That's because it only ever called itself, which is nonsensical.  That just means that all classes that derive from it must implement it.
2.  Make clang aware that one of the tests here just wants to test that equality doesn't throw, so we don't care about the result.

### DIFF
--- a/include/message_filters/subscriber.h
+++ b/include/message_filters/subscriber.h
@@ -102,16 +102,14 @@ public:
    *
    * \param node The rclcpp::Node to use to subscribe.
    * \param topic The topic to subscribe to.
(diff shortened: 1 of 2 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19832/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14330/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20554/)

ahcorde (approved):
LGTM with green CI

PR #106, opened 2023-12-07
Remove 'using' keyword in message_filters

clalancette:
The 'using' keyword can cause problems if multiple APIs have the same named functions in different namespaces.  Further, we tend not to use it very much in ROS 2.

Thus, this PR just removes the keyword 'using' from all of the tests in message_filters.  There is also a minor amount of style cleanup; a follow-up PR will go deeper into that to attempt to get our linters running here.

### DIFF
--- a/test/msg_cache_unittest.cpp
+++ b/test/msg_cache_unittest.cpp
@@ -42,8 +42,6 @@
 #include "message_filters/cache.h"
 #include "message_filters/message_traits.h"
 
-using namespace message_filters ;
-
 struct Header
 {
   rclcpp::Time stamp ;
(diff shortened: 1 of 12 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19960/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14430/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20665/)

test/msg_cache_unittest.cpp
```diff
@@ -70,27 +69,24 @@ struct TimeStamp<Msg>
...
 
 TEST(Cache, easyInterval)
 {
-  Cache<Msg> cache(10) ;
-  fillCacheEasy(cache, 0, 5) ;
+  message_filters::Cache<Msg> cache(10);
+  fillCacheEasy(cache, 0, 5);
 
-  vector<std::shared_ptr<Msg const> > interval_data = cache.getInterval(rclcpp::Time(5, 0), rclcpp::Time(35, 0)) ;
+  std::vector<std::shared_ptr<Msg const> > interval_data = cache.getInterval(rclcpp::Time(5, 0), rclcpp::Time(35, 0)) ;
```
fujitatomoya, line 89:
```suggestion
  std::vector<std::shared_ptr<Msg const> > interval_data = cache.getInterval(rclcpp::Time(5, 0), rclcpp::Time(35, 0));
```

test/msg_cache_unittest.cpp
```diff
@@ -147,15 +143,15 @@ std::shared_ptr<Msg const> buildMsg(int32_t seconds, int data)
...
+  message_filters::Cache<Msg> cache(10);
 
   cache.add(buildMsg(10, 1)) ;
   cache.add(buildMsg(30, 3)) ;
   cache.add(buildMsg(70, 7)) ;
   cache.add(buildMsg( 5, 0)) ;
   cache.add(buildMsg(20, 2)) ;
 
-  vector<std::shared_ptr<Msg const> > interval_data = cache.getInterval(rclcpp::Time(3, 0), rclcpp::Time(15, 0)) ;
+  std::vector<std::shared_ptr<Msg const> > interval_data = cache.getInterval(rclcpp::Time(3, 0), rclcpp::Time(15, 0)) ;
```
fujitatomoya, line 154:
```suggestion
  std::vector<std::shared_ptr<Msg const> > interval_data = cache.getInterval(rclcpp::Time(3, 0), rclcpp::Time(15, 0));
```

fujitatomoya (changes requested):
out of curiosity, are we not supposed to use `using`? what is the motivation for this change?

clalancette:
> out of curiosity, are we not supposed to use `using`? what is the motivation for this change?

It can cause subtle issues, particularly in header files.  The canonical example is doing something like this in a header file:

```
using std::placeholders::_1;
..

x = std::bind(&func, _1);
```

If another project wanted to `#include` your header file, but they did something like:

```
using boost::placeholders::_1;
```

Then it could cause things to break.  Also, the Google style guide says not to use it: https://google.github.io/styleguide/cppguide.html#Namespaces , because it can "pollute" the namespace.

clalancette:
Since the changes I pushed were only to spacing, I'm going to rely on the Rpr job here and just go ahead and merge this.  Thanks for the review!

Issue #110, opened 2024-01-21
Is there a bug in Latest time policy rates_ initialization?


wuchenhaogit:
In include/message_filters/sync_policies/latest_time.h

rates_ is array of struct Rate. I thought that its index corresponds to channel index. But its initialization in function initialize_rate<i>() doesn't put the struct (including alphas and prev) to index "i". Instead "push_back" is used. As a result,  the index of rates_ seems to represent the order of arrival of the first message of each channel. This appears confusing to me. How can  #I properly set the alphas for each channel?

Here is the source code of initialize_rate():
  template<int i>
  void initialize_rate()
  {
    if (rate_configs_.size() > 0U) {
      double rate_ema_alpha{Rate::DEFAULT_RATE_EMA_ALPHA};
      double error_ema_alpha{Rate::DEFAULT_ERROR_EMA_ALPHA};
      double rate_step_change_margin_factor{Rate::DEFAULT_MARGIN_FACTOR};
      if (rate_configs_.size() == RealTypeCount::value) {
        std::tie (
          rate_ema_alpha,
          error_ema_alpha,
          rate_step_change_margin_factor) = rate_configs_[i];
      } else if (rate_configs_.size() == 1U) {
        std::tie (
          rate_ema_alpha,
          error_ema_alpha,
          rate_step_change_margin_factor) = rate_configs_[0U];
      }
      rates_.push_back(
        Rate(
          ros_clock_->now(),
          rate_ema_alpha,
          error_ema_alpha,
          rate_step_change_margin_factor));
    } else {
      rates_.push_back(Rate(ros_clock_->now()));
    }
  }

andermi:
Confirmed.  I'll submit a PR. Thanks for finding this!

PR #111, opened 2024-01-23
rates_ vector out of channel order

andermi:
`rates_` vector out of channel order when user-specified `RateConfig` vector provided

see https://github.com/ros2/message_filters/issues/110

### DIFF
--- a/include/message_filters/sync_policies/latest_time.h
+++ b/include/message_filters/sync_policies/latest_time.h
@@ -49,7 +49,7 @@ Synchronizer<latest_policy> sync_policies(latest_policy(), caminfo_sub, limage_s
 sync_policies.registerCallback(callback);
 \endverbatim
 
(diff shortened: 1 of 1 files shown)


clalancette:
I'm honestly not sure that this is the correct fix.  From what I can tell of the current code, it is actually possible (if non-intuitive) to configure the rates properly.  You just need to call `setRateConfigPerMessage` of the correct size, and then all of the calls to `add` will reference the correct configuration as they are created.  And then `sort_indices` will actually sort them by their expiry time.

However, I'm not 100% sure of all of that.  What I think we really need is an example/test that shows the problem that we are trying to solve, and the way we would solve it with the current code.  Then we can consider whether this proposal solves it better.

andermi:
Sounds good! I can try adding a test to replicate #110

PR #115, opened 2024-02-28
Update TimeSynchronizer usage example.

rkeating-planted:
Update TimeSynchronizer usage example.


clalancette (approved):
Thanks for the fix.

PR #119, opened 2024-05-04
Move Docs From Wiki

CursedRock17:
As per the title, this pull request moves the documentation of `message_filters` over to `docs.ros.org` from the WIKI, fully updated to fit current ROS 2 stadards. Furthermore, this pr contians new, in-depth tutorials for users to get better with `message_filters` and understand the various filters and policies. That part of the pull request comes from both [this `ros2_documentation` issue](https://github.com/ros2/ros2_documentation/issues/3218) and [this `ros2_documentation` issue](https://github.com/ros2/ros2_documentation/issues/2272) in which the tutorials needed an update. So, I added those tutorials over here so the docs would be centralized and only people using the package would have to worry about it, but if we needed, we can just copy all the information over to a PR in that repo. Additionally, I cannot get the ros2 docs to work for Python, though it works in C++, it comes from this line: https://github.com/CursedRock17/message_filters/blob/66c6bf9230c95cecc87c974500d759772a4e16de/doc/index.rst?plain=1#L263


doc/index.rst
```diff
@@ -0,0 +1,268 @@
+Message Filters
+===============
+
+:mod:`message_filters` is a collection of message "filters" which take messages in,
+either from a ROS 2 subscription or another filter,
+and may or may not output the message
+at some time in the future, depending on a policy defined for that filter.
+
+message_filters also defines a common interface for these filters, allowing you to chain them together.
```
ahcorde, line 9:
```suggestion
``message_filters`` also defines a common interface for these filters, allowing you to chain them together.
```

doc/index.rst
```diff
@@ -0,0 +1,268 @@
...
+-------------
+The Subscriber filter is simply a wrapper around a ROS 2 subscription that provides a source for other filters. The Subscriber filter cannot connect to another filter's output, instead it uses a ROS 2 topic as its input.
+
+2.1 Connections
+~~~~~~~~~~~~~~~
+Input:
+  No input connections
+Output:
+  * C++: void callback(const std::shared_ptr<M const>&)
+  * Python: callback(msg)
```
ahcorde, line 52:
```suggestion
  * C++: ``void callback(const std::shared_ptr<M const>&)``
  * Python: ``callback(msg)``
```

doc/index.rst
```diff
@@ -0,0 +1,268 @@
...
+
+3.1 Connections
+~~~~~~~~~~~~~~~
+Input:
+  * C++: Up to 9 separate filters, each of which is of the form void callback(const std::shared_ptr<M const>&). The number of filters supported is determined by the number of template arguments the class was created with.
+  * Python: N separate filters, each of which has signature callback(msg).
+
+Output:
+  * C++: For message types M0..M8, void callback(const std::shared_ptr<M0 const>&, ..., const std::shared_ptr<M8 const>&). The number of parameters is determined by the number of template arguments the class was created with.
+  * Python: callback(msg0.. msgN). The number of parameters is determined by the number of template arguments the class was created with.
```
ahcorde, line 87:
```suggestion
  * Python: ``callback(msg0.. msgN)``. The number of parameters is determined by the number of template arguments the class was created with.
```

doc/index.rst
```diff
@@ -0,0 +1,268 @@
...
+4. Time Sequencer
+-----------------
+* Python: the TimeSequencer filter is not yet implemented.
+
+The TimeSequencer filter guarantees that messages will be called in temporal order according to their header's timestamp. The TimeSequencer is constructed with a specific delay which specifies how long to queue up messages before passing them through. A callback for a message is never invoked until the messages' time stamp is out of date by at least delay. However, for all messages which are out of date by at least the delay, their callback are invoked and guaranteed to be in temporal order. If a message arrives from a time prior to a message which has already had its callback invoked, it is thrown away.
+
+4.1 Connections
+~~~~~~~~~~~~~~~
+Input:
+  * C++: void callback(const std::shared_ptr<M const>&)
```
ahcorde, line 98:
```suggestion
  * C++: ``void callback(const std::shared_ptr<M const>&)``
```

doc/index.rst
```diff
@@ -0,0 +1,268 @@
...
+Output:
+  * C++: void callback(const std::shared_ptr<M const>&)
```
ahcorde, line 100:
```suggestion
  * C++: ``void callback(const std::shared_ptr<M const>&)``
```

doc/index.rst
```diff
@@ -0,0 +1,268 @@
...
+The Cache immediately passes messages through to its output connections.
+
+5.1 Connections
+~~~~~~~~~~~~~~~
+Input:
+  * C++: void callback(const std::shared_ptr<M const>&)
+  * Python: callback(msg)
+Output:
+  * C++: void callback(const std::shared_ptr<M const>&)
+  * Python: callback(msg)
```
ahcorde, line 128:
```suggestion
  * C++: ``void callback(const std::shared_ptr<M const>&)``
  * Python: ``callback(msg)``
Output:
  * C++: ``void callback(const std::shared_ptr<M const>&)``
  * Python: ``callback(msg)``
```

doc/index.rst
```diff
@@ -0,0 +1,268 @@
...
+
+5.3 Example (Python)
+~~~~~~~~~~~~~~~~~~~~
+.. code-block:: Python
+
+    sub = message_filters.Subscriber('my_topic', sensor_msgs.msg.Image)
+    cache = message_filters.Cache(sub, 100)
+
+In this example, the Cache stores the last 100 messages received on my_topic, and myCallback is called on the addition of every new message. The user can then make calls like cache.getInterval(start, end) to extract part of the cache.
+If the message type does not contain a header field that is normally used to determine its timestamp, and the Cache is contructed with allow_headerless=True, the current ROS 2 time is used as the timestamp of the message. This is currently only available in Python.
```
ahcorde, line 146:
```suggestion
In this example, the Cache stores the last 100 messages received on ``my_topic``, and ``myCallback`` is called on the addition of every new message. The user can then make calls like ``cache.getInterval(start, end)`` to extract part of the cache.
If the message type does not contain a header field that is normally used to determine its timestamp, and the Cache is contructed with ``allow_headerless=True``, the current ROS 2 time is used as the timestamp of the message. This is currently only available in Python.
```

doc/index.rst
```diff
@@ -0,0 +1,268 @@
...
+
+6. PolicyBased Synchronizers
+----------------------------
+The Synchronizer filter synchronizes incoming channels by the timestamps contained in their headers, and outputs them in the form of a single callback that takes the same number of channels. The C++ implementation can synchronize up to 9 channels.
+
+The Synchronizer filter is templated on a policy that determines how to synchronize the channels. There are currently two policies: ExactTime and ApproximateTime.
```
ahcorde, line 152: There is a third one [ApproximateEpsilonTime](https://github.com/ros2/message_filters/blob/rolling/include/message_filters/sync_policies/approximate_epsilon_time.h). Not need to write a independent tutorial in C++ and Python, you can open an issue with this task

doc/index.rst
```diff
@@ -0,0 +1,268 @@
...
+
+6.1 Connections
+~~~~~~~~~~~~~~~
+Input:
+  * C++: Up to 9 separate filters, each of which is of the form void callback(const std::shared_ptr<M const>&). The number of filters supported is determined by the number of template arguments the class was created with.
+  * Python: N separate filters, each of which has signature callback(msg).
+Output:
+  * C++: For message types M0..M8, void callback(const std::shared_ptr<M0 const>&, ..., const std::shared_ptr<M8 const>&). The number of parameters is determined by the number of template arguments the class was created with.
+  * Python: callback(msg0.. msgN). The number of parameters is determined by the number of template arguments the class was created with.
```
ahcorde, line 161:
```suggestion
  * C++: Up to 9 separate filters, each of which is of the form ``void callback(const std::shared_ptr<M const>&)``. The number of filters supported is determined by the number of template arguments the class was created with.
  * Python: N separate filters, each of which has signature ``callback(msg)``.
Output:
  * C++: For message types M0..M8, ``void callback(const std::shared_ptr<M0 const>&, ..., const std::shared_ptr<M8 const>&)``. The number of parameters is determined by the number of template arguments the class was created with.
  * Python: ``callback(msg0.. msgN)``. The number of parameters is determined by the number of template arguments the class was created with.
```

doc/index.rst
```diff
@@ -0,0 +1,268 @@
...
+
+6.2 ExactTime Policy
+~~~~~~~~~~~~~~~~~~~~
+The message_filters::sync_policies::ExactTime policy requires messages to have exactly the same timestamp in order to match. Your callback is only called if a message has been received on all specified channels with the same exact timestamp. The timestamp is read from the header field of all messages (which is required for this policy).
```
ahcorde, line 165:
```suggestion
The ``message_filters::sync_policies::ExactTime`` policy requires messages to have exactly the same timestamp in order to match. Your callback is only called if a message has been received on all specified channels with the same exact timestamp. The timestamp is read from the header field of all messages (which is required for this policy).
```

doc/index.rst
```diff
@@ -0,0 +1,268 @@
...
+
+6.3 ApproximateTime Policy
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+The message_filters::sync_policies::ApproximateTime policy uses an adaptive algorithm to match messages based on their timestamp.
+
+If not all messages have a header field from which the timestamp could be determined, see below for a workaround. If some messages are of a type that doesn't contain the header field, ApproximateTimeSynchronizer refuses by default adding such messages. However, its Python version can be constructed with allow_headerless=True, which uses current ROS 2 time in place of any missing header.stamp field:
```
ahcorde, line 171:
```suggestion
The ``message_filters::sync_policies::ApproximateTime`` policy uses an adaptive algorithm to match messages based on their timestamp.

If not all messages have a header field from which the timestamp could be determined, see below for a workaround. If some messages are of a type that doesn't contain the header field, ``ApproximateTimeSynchronizer`` refuses by default adding such messages. However, its Python version can be constructed with ``allow_headerless=True``, which uses current ROS 2 time in place of any missing header.stamp field:
```

doc/index.rst
```diff
@@ -0,0 +1,268 @@
...
+The Chain filter allows you to dynamically chain together multiple singleinput/single-output (simple) filters. As filters are added to it they are automatically connected together in the order they were added. It also allows you to retrieve added filters by index.
+
+Chain is most useful for cases where you want to determine which filters to apply at runtime rather than compiletime.
+
+7.1 Connections
+~~~~~~~~~~~~~~~
+Input:
+  * C++: void callback(const std::shared_ptr<M const>&)
+Output:
+  * C++: void callback(const std::shared_ptr<M const>&)
```
ahcorde, line 185:
```suggestion
Input:
  * C++: ``void callback(const std::shared_ptr<M const>&)``
Output:
  * C++: ``void callback(const std::shared_ptr<M const>&)``
```

doc/index.rst
```diff
@@ -0,0 +1,268 @@
...
+where ``msg`` is the message.
+
+Message filters that accept input from an upstream
+message filter (e.g. :class:`message_filters.Cache`) register their own
+message handler as a callback.
+
+Output connections are registered through the ``registerCallback()`` function.
+
+
+Here's a simple example of using a Subscriber with a Cache::
```
ahcorde, line 242:
```suggestion
Output connections are registered through the ``registerCallback()`` function.

Here's a simple example of using a Subscriber with a Cache::
```

doc/index.rst
```diff
@@ -0,0 +1,268 @@
...
+
+    def myCallback(posemsg):
+       print(posemsg)
+
+    sub = message_filters.Subscriber(node_object, robot_msgs.msg.Pose, "pose_topic")
+    cache = message_filters.Cache(sub, 10)
+    cache.registerCallback(myCallback)
+
+The Subscriber here acts as the source of messages.  Each message is passed to the cache, which then passes it through to the
+user's callback ``myCallback``.
```
ahcorde, line 252:
```suggestion
The Subscriber here acts as the source of messages. Each message is passed to the cache, which then passes it through to the
user's callback ``myCallback``.
```

doc/Tutorials/Approximate-Synchronizer-Python.rst
```diff
@@ -0,0 +1,155 @@
...
+`sensor_msgs <https://github.com/ros2/common_interfaces/tree/rolling/sensor_msgs/msg>`_.
+To simulate a working ``ApproximateTimeSynchronizer``. We will be publishing and subscribing to topics of those respective types, to showcase how real sensors would be working. To simulate them we will also need two ``Timers`` on different intervals. Then, we will be utilizing an ``ApproximateTimeSynchronizer`` to get these messages from the sensor topics aligned with a slight delay between messages..
+
+
+It is essential that the QoS is the same for all of the publishers and subscribers, otherwise the Message Filter cannot align the topics together. So, create one ``QoSProfile`` and stick with it, or find out what ``qos`` is being used in the native sensor code, and replicate it. Do basic construction of each object relating to the ``Node`` and callback methods that may be used in the future. Both of the two timers we utilize will have different timer values of ``1`` and ``1.05`` which causes the timers to off at different points, which is an advantage of using ``ApproximateTime``. Notice that we must call ``sync->registerCallback`` to sync up the two (or more) chosen topics.
```
ahcorde, line 53:
```suggestion
To simulate a working ``ApproximateTimeSynchronizer``. We will be publishing and subscribing to topics of those respective types, to showcase how real sensors would be working. To simulate them we will also need two ``Timers`` on different intervals. Then, we will be utilizing an ``ApproximateTimeSynchronizer`` to get these messages from the sensor topics aligned with a slight delay between messages..

It is essential that the QoS is the same for all of the publishers and subscribers, otherwise the Message Filter cannot align the topics together. So, create one ``QoSProfile`` and stick with it, or find out what ``qos`` is being used in the native sensor code, and replicate it. Do basic construction of each object relating to the ``Node`` and callback methods that may be used in the future. Both of the two timers we utilize will have different timer values of ``1`` and ``1.05`` which causes the timers to off at different points, which is an advantage of using ``ApproximateTime``. Notice that we must call ``sync->registerCallback`` to sync up the two (or more) chosen topics.
```

doc/Tutorials/Approximate-Synchronizer-Python.rst
```diff
@@ -0,0 +1,155 @@
...
+
+   'approximate_time_sync = pkg_name.approximate_time_sync:main',
+
+
+3. Build
+~~~~~~~~
+From the root of your package, build and source.
+
+
+.. code-block:: bash
```
ahcorde, line 138:
```suggestion
From the root of your package, build and source.

.. code-block:: bash
```

doc/Tutorials/Writing-A-Time-Synchronizer-Cpp.rst
```diff
@@ -0,0 +1,191 @@
...
+        time_sync
+        DESTINATION lib/${PROJECT_NAME})
+
+
+3. Build
+~~~~~~~~
+From the root of your package, build and source.
+
+
+.. code-block:: bash
```
ahcorde, line 175:
```suggestion
From the root of your package, build and source.

.. code-block:: bash
```

ahcorde (changes requested):

CursedRock17:
I've added those necessary changes but when it comes to [this comment](https://github.com/ros2/message_filters/pull/119#discussion_r1604565909), we're looking to do a whole other Pull request for those tutorials after this one gets merged correct? Also, should I add a section for the `latest_time` policy as well?

ahcorde (approved):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21110/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15409/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21824/)

ahcorde (changes requested):
I can see these errors:

```bash
/home/ahcorde/ros2_rolling/src/ros2/message_filters/doc/Tutorials/Approximate-Synchronizer-Cpp.rst:1: ERROR: Unknown directive type "redirect-from".

.. redirect-from::

   Tutorials/Approximate-Synchronizer-Cpp
/home/ahcorde/ros2_rolling/src/ros2/message_filters/doc/Tutorials/Approximate-Synchronizer-Python.rst:1: ERROR: Unknown directive type "redirect-from".

.. redirect-from::

   Tutorials/Approximate-Synchronizer-Python
/home/ahcorde/ros2_rolling/src/ros2/message_filters/doc/Tutorials/Writing-A-Time-Synchronizer-Cpp.rst:1: ERROR: Unknown directive type "redirect-from".

.. redirect-from::

   Tutorials/Writing-A-Time_Synchronizer-Cpp
/home/ahcorde/ros2_rolling/src/ros2/message_filters/doc/Tutorials/Writing-A-Time-Synchronizer-Python.rst:1: ERROR: Unknown directive type "redirect-from".

.. redirect-from::

   Tutorials/Writing-A-Time_Synchronizer-Python
```

In the `index.html` I can't see any link to the tutorials or API links

CursedRock17:
Yeah I messed up the `index.rst` file when I was dealing with the merge conflict, added in an extra one somehow. Turns out those redirects were useless so I removed those and created a singular, correct index file. Though now I've got all the updated code in the pull request, so I assume I need to remove in order to have a nice PR.

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21168/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15473/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21892/)

CursedRock17:
I don't know what I had done before with my rebase but it completely messed up the last few commits, all the linters pass now and we shouldn't have any CI problems, additionally all the tutorial links work

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21170/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15475/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21894/)

PR #125, opened 2024-05-22
Adding cpplint

CursedRock17:
This pull request is part of a group of pull request which break apart #120, this particular pull request adds `cpplint` to `message_filters`.


include/message_filters/message_event.h
```diff
@@ -43,7 +45,7 @@
 #define RCUTILS_ASSERT assert
 #endif
 // Uncomment below intead
-//#include <rcutils/assert.h>
+// #include <rcutils/assert.h>
```
clalancette, line 48: Let's just remove this (and the comment above).

include/message_filters/parameter_adapter.h
```diff
@@ -79,7 +79,7 @@ struct ParameterAdapter
     return *event.getMessage();
   }
 };
-//struct message_filters::ParameterAdapter<const std::shared_ptr<const Msg> &>
+// struct message_filters::ParameterAdapter<const std::shared_ptr<const Msg> &>
```
clalancette, line 82: Just remove this.

include/message_filters/sync_policies/approximate_time.h
```diff
@@ -59,7 +60,7 @@
 #define RCUTILS_BREAK std::abort
 #endif
 // Uncomment below intead
-//#include <rcutils/assert.h>
+// #include <rcutils/assert.h>
```
clalancette, line 63: Just remove this (and the comment above).

include/message_filters/sync_policies/approximate_time.h
```diff
@@ -368,9 +376,9 @@ struct ApproximateTime : public PolicyBase<M0, M1, M2, M3, M4, M5, M6, M7, M8>
 
   void makeCandidate()
   {
-    //printf("Creating candidate\n");
+    // printf("Creating candidate\n");
```
clalancette, line 379: Just remove the line.

include/message_filters/sync_policies/approximate_time.h
```diff
@@ -404,7 +412,7 @@ struct ApproximateTime : public PolicyBase<M0, M1, M2, M3, M4, M5, M6, M7, M8>
     std::get<6>(past_).clear();
     std::get<7>(past_).clear();
     std::get<8>(past_).clear();
-    //printf("Candidate created\n");
+    // printf("Candidate created\n");
```
clalancette, line 415: Remove the line.

include/message_filters/sync_policies/approximate_time.h
```diff
@@ -476,7 +484,7 @@ struct ApproximateTime : public PolicyBase<M0, M1, M2, M3, M4, M5, M6, M7, M8>
   // Assumes: all deques are non empty, i.e. num_non_empty_deques_ == RealTypeCount::value
   void publishCandidate()
   {
-    //printf("Publishing candidate\n");
+    // printf("Publishing candidate\n");
```
clalancette, line 487: Remove the line.

include/message_filters/sync_policies/approximate_time.h
```diff
@@ -661,9 +669,9 @@ struct ApproximateTime : public PolicyBase<M0, M1, M2, M3, M4, M5, M6, M7, M8>
     // While no deque is empty
     while (num_non_empty_deques_ == (uint32_t)RealTypeCount::value) {
       // Find the start and end of the current interval
-      //printf("Entering while loop in this state [\n");
-      //show_internal_state();
-      //printf("]\n");
+      // printf("Entering while loop in this state [\n");
+      // show_internal_state();
+      // printf("]\n");
```
clalancette, line 674: Remove all of these lines.

include/message_filters/sync_policies/approximate_time.h
```diff
@@ -714,8 +722,8 @@ struct ApproximateTime : public PolicyBase<M0, M1, M2, M3, M4, M5, M6, M7, M8>
       }
       // INVARIANT: we have a candidate and pivot
       RCUTILS_ASSERT(pivot_ != NO_PIVOT);
-      //printf("start_index == %d, pivot_ == %d\n", start_index, pivot_);
-      if (start_index == pivot_) {  // TODO: replace with start_time == pivot_time_
+      // printf("start_index == %d, pivot_ == %d\n", start_index, pivot_);
```
clalancette, line 725: Remove the line.

include/message_filters/signal9.h
```diff
@@ -175,7 +175,7 @@ class Signal9
     typename P6, typename P7, typename P8>
   Connection addCallback(const std::function<void(P0, P1, P2, P3, P4, P5, P6, P7, P8)> & callback)
   {
-    CallbackHelper9T<P0, P1, P2, P3, P4, P5, P6, P7, P8> * helper =
+    explicit CallbackHelper9T<P0, P1, P2, P3, P4, P5, P6, P7, P8> * helper =
```
clalancette, line 178:
```suggestion
    CallbackHelper9T<P0, P1, P2, P3, P4, P5, P6, P7, P8> * helper =
```

include/message_filters/signal9.h
```diff
@@ -107,7 +107,7 @@ class CallbackHelper9T
       typename A5::Parameter, typename A6::Parameter, typename A7::Parameter,
       typename A8::Parameter)> Callback;
 
-  explicit CallbackHelper9T(const Callback & cb)
+  CallbackHelper9T(const Callback & cb)
```
clalancette, line 110: Why remove `explicit` here?

clalancette (changes requested):

CursedRock17:
So I've fixed the issues with `cpplint` though there are concerns with the `explicit` constructors. 
1. When it comes to `MessageEvent`, there was a call to [make its constructor explicit](https://github.com/ros2/message_filters/blob/rolling/include/message_filters/message_event.h#L119) in ROS 2, which is fine, but it means callbacks must follow the type of `const ConstMessagePtr &`. Thus, [this line](https://github.com/ros2/message_filters/blob/86e8c0c7fa5909447b963e50270edc5b2759074b/test/test_simple.cpp#L148) was changed to use the correct callbacks, creating errors otherwise. These changes now result in an API break, so is there a way to silence the explication warnings for this, should I create explicit constructors for each message type, or should we move on regardless?
2. There are also issues with the `explicit` keyword within `sync_policies`, in which the `Policy` must be constructed and then passed to the `Synchronizer` like [here](https://github.com/ros2/message_filters/blob/b8dbbf865506632139c9ea7202ea5a95e5cdaee2/test/test_exact_time_policy.cpp#L121-L122). Not a big deal work wise, but technically this will also cause an API break, which isn't a correct practice to follow, so should we also silence these errors instead of creating an `explicit` constructor within the policies?

clalancette:
> So I've fixed the issues with `cpplint` though there are concerns with the `explicit` constructors.

So technically, marking *any* of these constructors `explicit` is an API break; things that compiled before may no longer compile.

While that isn't necessarily a bad thing, we should probably decouple it from this PR so we can get this in.  In that case, what I'm going to suggest is that everywhere cpplint complains about explicit constructors, we add a `// NOLINT(explicit)` to the line.  Once we get this PR in, we can consider whether we actually want to do that API break separately.  Does that sound reasonable?

ahcorde (approved):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21203/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15513/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21925/)

clalancette (approved):
These changes look great to me with green CI.

@CursedRock17 Can I suggest that you open an issue to follow-up about the `NOLINT(runtime/explicit)` comments?

PR #129, opened 2024-06-07
Adding explicit constructors

### LINKED ISSUE #128: Improve Explicit Constructors
As part of pull request #125, `cpplint` suggests we mark all single param constructors as `explicit` as it is a good practice for writing c++ code. Technically marking constructors as `explicit` constitutes an API break, so to prevent `cpplint` from going off they're marked as `NOLINT`. An example of an API break from these constructors is with [`Synchronizer`](https://github.com/ros2/message_filters/blob/rolling/include/message_filters/synchronizer.h#L207), in which the `Policy` must be created and initialized [first](https://github.com/ros2/message_filters/blob/b8dbbf865506632139c9ea7202ea5a95e5cdaee2/test/test_exact_time_policy.cpp#L121-L122), while much of the community just passes `queue_size` straight to their constructor. We should find a way to make these constructors explicit, while preventing API breaks

CursedRock17:
This pull request is an extension of #120 meant to resolve #128 in which we restore the ability to use the `explicit` keyword for singular parameter constructors. The current PR should work for `message_filters` but since making constructors `explicit` will technically cause an API break, I'm also testing downstream repositories. I will shortly display a list of all places where these changes cause a break, from there we can determine the extent in which we need to act

### DIFF
--- a/include/message_filters/message_event.h
+++ b/include/message_filters/message_event.h
@@ -118,7 +118,7 @@ public:
   /**
    * \todo Make this explicit in ROS 2.0.  Keep as auto-converting for now to maintain backwards compatibility in some places (message_filters)
    */
-  MessageEvent(const ConstMessagePtr & message)  // NOLINT(runtime/explicit)
+  explicit MessageEvent(const ConstMessagePtr & message)
   {
     init(message, rclcpp::Clock().now(), true, message_filters::DefaultMessageCreator<Message>());
   }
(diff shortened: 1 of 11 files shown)


ahcorde (approved):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21238/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15556/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/715/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21960/)

PR #132, opened 2024-06-18
Revert "Adding explicit constructors (#129)"

clalancette:
This reverts commit aea0e9ffcdeef43c1244a413a1f5ac2ffb78d082.

This is causing downstream packages to fail to compile; see for example https://build.ros2.org/view/Rbin_uN64/job/Rbin_uN64__pcl_ros__ubuntu_noble_amd64__binary/23/console .  I verified that removing this commit causes those packages to compile again.  If we want to do this, we may have to be a little more careful on how this is done.

@CursedRock17 @ahcorde FYI.

### DIFF
--- a/include/message_filters/message_event.h
+++ b/include/message_filters/message_event.h
@@ -118,7 +118,7 @@ public:
   /**
    * \todo Make this explicit in ROS 2.0.  Keep as auto-converting for now to maintain backwards compatibility in some places (message_filters)
    */
-  explicit MessageEvent(const ConstMessagePtr & message)
+  MessageEvent(const ConstMessagePtr & message)  // NOLINT(runtime/explicit)
   {
     init(message, rclcpp::Clock().now(), true, message_filters::DefaultMessageCreator<Message>());
   }
--- a/include/message_filters/pass_through.h
+++ b/include/message_filters/pass_through.h
@@ -53,7 +53,7 @@ public:
 
 
   template<typename F>
-  explicit PassThrough(F & f)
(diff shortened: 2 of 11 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21282/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15600/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/778/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22017/)

ahcorde (approved):

ahcorde:
I almost have a patch for this packages, i will open it today. Just in case you want to wait

ahcorde:
PR open https://github.com/ros-perception/perception_pcl/pull/457

ahcorde:
@clalancette I can open any other PR, if there are more packages failing

clalancette:
> @clalancette I can open any other PR, if there are more packages failing

Hm.  Well, the fact that we broke downstream packages makes me inclined to revert this, even if we can fix the downstreams.  At the very least we should start by deprecating this if we want to change the API like this.

But let's see what the maintainers in https://github.com/ros-perception/perception_pcl/pull/457 have to say first.

clalancette:
I discussed this with @ahcorde , and we agreed for now to merge this in to get Rolling regression free.  We'll eventually follow-up with this and https://github.com/ros-perception/perception_pcl/pull/457 again in the future.

PR #195, opened 2025-07-02
cache.hpp: Add allow_headerless

### LINKED ISSUE #74: C++ equivalent of `allow_headerless` synchronization
I know that you can set `allow_headerless = True` in python, but is there an equivalent option for C++?  Are there plans to add it in the future?

EsipovPA:
## Description

Added allow_headerless parameter to CPP Cahce filter. Behaves the same way as the python implementation

Fixes #74 

If `allow_headerless` is set to `true`, the message filter allows to cache messages with no header, such as `std_msgs/msg/String`.
By default, `allow_headerless` is set to `false`

No.

### DIFF
--- a/include/message_filters/cache.hpp
+++ b/include/message_filters/cache.hpp
@@ -31,8 +31,9 @@
 
 #include <cstddef>
 #include <deque>
-#include <memory>
 #include <functional>
+#include <memory>
+#include <stdexcept>
 #include <vector>
 
 #include <rclcpp/rclcpp.hpp>
@@ -67,9 +68,21 @@ public:
 
   template<class F>
   explicit Cache(F & f, unsigned int cache_size = 1)
+  : Cache(f, cache_size, false) {}
+
+  template<class F>
+  explicit Cache(F & f, unsigned int cache_size, bool allow_headerless)
   {
     setCacheSize(cache_size);
     connectInput(f);
+
+    if (message_filters::message_traits::HasHeader<M>::value) {
+      getMessageTime = getMessageTimeFromHeader;
+    } else if (allow_headerless) {
+      getMessageTime = getMessageReceiveTime;
+    } else {
+      throw std::runtime_error("Caching messages with no header not allowed");
+    }
   }
 
   /**
@@ -78,8 +91,24 @@ public:
    * called later
    */
   explicit Cache(unsigned int cache_size = 1)
+  : Cache(cache_size, false) {}
+
+  /**
+   * Initializes a Message Cache without specifying a parent filter. This implies that in
+   * order to populate the cache, the user then has to call add themselves, or connectInput() is
+   * called later
+   */
+  explicit Cache(unsigned int cache_size, bool allow_headerless)
   {
     setCacheSize(cache_size);
+
+    if (message_filters::message_traits::HasHeader<M>::value) {
+      getMessageTime = getMessageTimeFromHeader;
+    } else if (allow_headerless) {
+      getMessageTime = getMessageReceiveTime;
+    } else {
+      throw std::runtime_error("Caching messages with no header not allowed");
+    }
   }
 
   template<class F>
@@ -144,10 +173,8 @@ public:
 
       // Keep walking backwards along deque until we hit the beginning,
       // or until we find a timestamp that's smaller than (or equal to) msg's timestamp
-      rclcpp::Time evt_stamp = mt::TimeStamp<M>::value(*evt.getMessage());
-      while (rev_it != cache_.rend() &&
-        mt::TimeStamp<M>::value(*(*rev_it).getMessage()) > evt_stamp)
-      {
+      rclcpp::Time evt_stamp = getMessageTime(evt);
+      while (rev_it != cache_.rend() && getMessageTime(*rev_it) > evt_stamp) {
         rev_it++;
       }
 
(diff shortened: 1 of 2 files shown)


test/msg_cache_unittest.cpp
```diff
@@ -85,29 +87,44 @@ TEST(Cache, emptySurroundingInterval)
...
 {
   message_filters::Cache<Msg> cache(10);
   fillCacheEasy(cache, 0, 5);
 
   std::vector<std::shared_ptr<Msg const>> interval_data = cache.getInterval(
-    rclcpp::Time(
-      5,
-      0), rclcpp::Time(
-      35, 0));
+    rclcpp::Time(5, 0, RCL_ROS_TIME),
```
mjcarroll, line 91: Are these changes necessary or are they just to be more explicit?

include/message_filters/cache.hpp
```diff
@@ -66,20 +66,37 @@ class Cache : public SimpleFilter<M>
   typedef MessageEvent<M const> EventType;
 
   template<class F>
-  explicit Cache(F & f, unsigned int cache_size = 1)
+  explicit Cache(F & f, unsigned int cache_size = 1, bool allow_headerless = false)
```
mjcarroll: I think this is fine from an API point of view, but I think there may be some ABI implications around adding a new parameter with a default value.

Any downstream consumers of message filters would need to recompile, which is generally not a problem, but means that this could not be backported.

EsipovPA: Maybe it is better for backward compatibility to add two overloaded constructors with this signature, and new functionality, and keep the old interface intact?

test/msg_cache_unittest.cpp
EsipovPA: I've encountered the need for this changes after switching to `std_msgs::Header` [here](https://github.com/ros2/message_filters/pull/195/files#diff-19378d388e472cc2268d17b12e8123dd313afce72b5efeb83ff4c4979befadeeL39-L42). Until I've introduced the specific time source here, I was having the following time comparison error:

```
2: [ RUN      ] Cache.easyInterval      
2: unknown file: Failure                 
2: C++ exception with description "can't compare times with different time sources" thrown in the test body.
2:                           
```

If I'm correct, [this TimeStamp function](https://github.com/ros2/message_filters/blob/rolling/include/message_filters/message_traits.hpp#L95-L102) being called instead of [this one](https://github.com/ros2/message_filters/blob/rolling/test/msg_cache_unittest.cpp#L55-L62). Actually, it seems like I should remove the last one, as it is not called now. But in the end, it seems to be the the behavior closer to the actual behavior of the Cache filterin the real life applications.

include/message_filters/cache.hpp
EsipovPA: I've tried to do exactly that, while removing default value from new constructors signatures. And it resulted in call ambiguity. Not sure about the better way to address this.

ahcorde: As @mjcarroll mentioned, If we want to backport this feature we need to create a new constructor. Otherwise we will break ABI compatibility.

You can find here a list of ABI breaking changes https://acodersjourney.com/20-abi-breaking-changes/

ahcorde (changes requested):

include/message_filters/cache.hpp
ahcorde: you can try with this new signature

```cpp
explicit Cache(F & f, unsigned int cache_size, bool allow_headerless)
```

include/message_filters/cache.hpp
```diff
@@ -67,9 +67,21 @@ class Cache : public SimpleFilter<M>
...
   {
     setCacheSize(cache_size);
     connectInput(f);
+
+    if (message_filters::message_traits::HasHeader<M>::value) {
+      getMessageTime = getMessageTimeFromHeader;
+    } else if (allow_headerless) {
+      getMessageTime = getMessageReceiveTime;
+    } else {
+      throw std::runtime_error("Caching messages with no header not allowed");
```
ahcorde, line 84: include `<stdexcept>`

ahcorde (changes requested):

include/message_filters/cache.hpp
EsipovPA: Done that. :slightly_smiling_face:

include/message_filters/cache.hpp
EsipovPA: Added new constructors. Seems like it is working now. With removed defaults it worked fine, with no call ambiguity. Am I correct that the current implementation will keep ABI intact?

ahcorde:
Pulls: ros2/message_filters#195
Gist: https://gist.githubusercontent.com/ahcorde/c35518380f5d26063ce59c4189bb79d3/raw/e9572e295789a79e62019a4081b1b32d88c0163c/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/16412
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/24206/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/18287/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/3639/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/24811/)

ahcorde (approved):

Issue #186, opened 2025-06-26
Better python filter-class interfaces


EsipovPA:
### Description

There are a few things, that seem like they may be done in a more python-way IMHO.

## Subscriber
Subscriber filter `__init__` signature looks like this:

```
def __init__(self, *args, **kwargs):
```
The *args are decomposed into positional arguments. 
```
    def __init__(self, *args, **kwargs):
        SimpleFilter.__init__(self)
        self.node = args[0]
        self.topic = args[2]
```

As a result, to initialize Subscriber filter, one has to pass node name, topic and message type as positional arguments. And the expected order of the positional arguments is not provided by the method signature, or by the class `help`. But the QoS profile can only be passed as a keyword argument. So the correct way to initialize the Subscriber filter looks like this:

```
            self.subscriber_filter = Subscriber(
                self,
                String,
                "/example/topic",
                qos_profile=qos_profile,
            )
```
It is the only filter in this library, that utilizes this approach. Others use named arguments. Maybe it is better to refactor the signature?

## Extremely short argument names and parameter names in come of the filter classes in general.

For example

```
    def __init__(self, fs, queue_size):
```
`fs` instead of `filters`, or `filters_list`. And some other examples. The `TimeSequencer` class seems to be in the best condition in that way. Maybe It will be better to refactor other classes with the same design philosophy in mind, as the `TimeSequencer` class? 

If nobody is working on this matter, I would like to try my best and provide PR with corresponding changes.

### Motivation

Better code readability. Simpler long term code support.

### Design / Implementation Considerations

Backwards compatibility.

ahcorde:
@EsipovPA 

Happy to review the PR if you decide to contribute

SamuelFoo:
Hi, I'm interested in submitting a PR as well if no one is working on this!

PR #220, opened 2025-10-31
Change function signature for Python Subscriber class

SamuelFoo:
## Description

A better function signature for the Python Subscriber class.

Related to #186 

No.

### Additional Information

### DIFF
--- a/src/message_filters/__init__.py
+++ b/src/message_filters/__init__.py
@@ -85,7 +85,7 @@ class Subscriber(SimpleFilter):
         topic: str,
         qos_profile: Union[QoSProfile, int] = 10,
         **kwargs,
-    ):
+    ) -> None:
         """
         Construct a Subscriber.
 
--- a/src/message_filters/__init__.py
+++ b/src/message_filters/__init__.py
@@ -83,7 +83,7 @@ class Subscriber(SimpleFilter):
         node: Node,
         msg_type: Type[MsgT],
         topic: str,
-        qos_profile: Union[QoSProfile, int]=10,
+        qos_profile: Union[QoSProfile, int] = 10,
         **kwargs,
     ):
         """
@@ -94,10 +94,11 @@ class Subscriber(SimpleFilter):
         node (Node): The node to create a subscriber for.
         msg_type (Type[MsgT]): The type of ROS messages the subscription will subscribe to.
         topic (str): The name of the topic the subscription will subscribe to.
-        qos_profile (QoSProfile | int): A QoSProfile or a history depth to apply to the subscription.
-            In the case that a history depth is provided, the QoS history is set to
-            KEEP_LAST, the QoS history depth is set to the value
-            of the parameter, and all other QoS settings are set to their default values.
+        qos_profile (QoSProfile | int): A QoSProfile or a history depth to apply to the
+            subscription. In the case that a history depth is provided, the QoS history is
+            set to KEEP_LAST, the QoS history depth is set to the value of the parameter,
+            and all other QoS settings are set to their default values.
+
         """
         SimpleFilter.__init__(self)
         self.node = node
--- a/src/message_filters/__init__.py
+++ b/src/message_filters/__init__.py
@@ -89,12 +89,10 @@ class Subscriber(SimpleFilter):
         """
         Construct a Subscriber.
 
-        Args:
-        ----
-        node (Node): The node to create a subscriber for.
-        msg_type (Type[MsgT]): The type of ROS messages the subscription will subscribe to.
-        topic (str): The name of the topic the subscription will subscribe to.
-        qos_profile (QoSProfile | int): A QoSProfile or a history depth to apply to the
+        :param node: The node to create a subscriber for.
+        :param msg_type: The type of ROS messages the subscription will subscribe to.
+        :param topic: The name of the topic the subscription will subscribe to.
+        :param qos_profile: A QoSProfile or a history depth to apply to the
             subscription. In the case that a history depth is provided, the QoS history is
             set to KEEP_LAST, the QoS history depth is set to the value of the parameter,
             and all other QoS settings are set to their default values.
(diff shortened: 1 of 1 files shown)


src/message_filters/__init__.py
```diff
@@ -76,12 +76,13 @@ class Subscriber(SimpleFilter):
     to it.
     """
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, node, msg_type, topic, qos_profile=10, **kwargs):
```
mjcarroll: @christophebedard and I talked about this.  There were two broken cases before:

1) users used _only_ kwargs (which would have thrown an IndexError right away)

2) Users passed more than 3 positional args, which would have caused a failure in node.create_subscription.

This should be perfectly fine backwards-compatibility wise if it was functional code before.

EsipovPA:
Thanks for contributing to this issue!

Though I woud suggest changing "Fixes #186" to "relates", or "partially fixes" so the issue is not automatically closed.

My motivation ts that this PR does not address the "Extremely short argument names and parameter names in some of the filter classes in general." section of the issue description.

SamuelFoo:
Got it! Made the change in the description.

src/message_filters/__init__.py
```diff
@@ -76,12 +78,34 @@ class Subscriber(SimpleFilter):
...
+
+        """
         SimpleFilter.__init__(self)
-        self.node = args[0]
-        self.topic = args[2]
-        kwargs.setdefault('qos_profile', 10)
-        self.sub = self.node.create_subscription(*args[1:], self.callback, **kwargs)
+        self.node = node
+        self.topic = topic
+        self.sub = self.node.create_subscription(
```
EsipovPA, line 103:
```suggestion
        self.subscription = self.node.create_subscription(
```
?

Do not insist. Just asking.

LGTM otherwise. @mjcarroll ?

EsipovPA (approved):

EsipovPA:
I would also suggest squashing commits so that only "Change __init__ signature..." commit remains. This will make the commit history clearer. Commit message "Fix style to pass CI tests" does not seem that informative in the long term, IMHO.

All the commits in this PR are related to this change anyway.

src/message_filters/__init__.py
```diff
@@ -76,12 +78,33 @@ class Subscriber(SimpleFilter):
...
 
-    def __init__(self, *args, **kwargs):
+    def __init__(
+        self,
+        node: Node,
+        msg_type: Type[MsgT],
+        topic: str,
+        qos_profile: Union[QoSProfile, int]=10,
+        **kwargs,
+    ):
```
christophebedard:
```suggestion
    ) -> None:
```

mjcarroll:
> I would also suggest squashing commits so that only "Change **init** signature..." commit remains. This will make the commit history clearer. Commit message "Fix style to pass CI tests" does not seem that informative in the long term, IMHO.
> 
> All the commits in this PR are related to this change anyway.

We generally suggest maintainers to squash merge, but appreciate it either way, as authors can summarize/describe changesets better in many cases

src/message_filters/__init__.py
```diff
@@ -76,12 +78,33 @@ class Subscriber(SimpleFilter):
     to it.
     """
 
-    def __init__(self, *args, **kwargs):
+    def __init__(
+        self,
+        node: Node,
+        msg_type: Type[MsgT],
+        topic: str,
+        qos_profile: Union[QoSProfile, int]=10,
```
christophebedard:
```suggestion
        qos_profile: Union[QoSProfile, int] = 10,
```

src/message_filters/__init__.py
```diff
@@ -76,12 +78,34 @@ class Subscriber(SimpleFilter):
...
+
+        Args:
+        ----
+        node (Node): The node to create a subscriber for.
+        msg_type (Type[MsgT]): The type of ROS messages the subscription will subscribe to.
+        topic (str): The name of the topic the subscription will subscribe to.
+        qos_profile (QoSProfile | int): A QoSProfile or a history depth to apply to the
+            subscription. In the case that a history depth is provided, the QoS history is
+            set to KEEP_LAST, the QoS history depth is set to the value of the parameter,
+            and all other QoS settings are set to their default values.
```
christophebedard: We don't have to include the param types in the docstring, since it's already typed

src/message_filters/__init__.py
```diff
@@ -76,12 +78,34 @@ class Subscriber(SimpleFilter):
...
+
```
christophebedard:
```suggestion
```

src/message_filters/__init__.py
christophebedard: That could break downstream users if they access the `sub` attribute, so I think we should keep `sub`.

EsipovPA: Makes sense.

Not being completely serious: Maybe add a property for backwards compatibility? :grin:

mjcarroll:
This just needs CI.  ci.ros2.org is sad this morning, but I will run it through once that clears up.

christophebedard:
Pulls: ros2/message_filters#220
Gist: https://gist.githubusercontent.com/christophebedard/6d572db60e44132e58e7a14358aa6648/raw/6d53f068be8b9ef6fe0d744e2d074a2423fefb8d/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/17387
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/25591/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/19622/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/5153/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/25828/)

christophebedard (approved):

mjcarroll:
Thanks @skyegalaxy for CI and thanks @SamuelFoo for coming out and hacking!

ahcorde (approved):

PR #238, opened 2025-12-17
#200 fix inconsistensy between cpp and python exact time synchronizer impl

### LINKED ISSUE #200: Inconsistency between C++ and Python implementation for ExactTimeSynchronizer
### Generated by Generative AI

_No response_

### Operating System:

```not relevant```

### ROS version or commit hash:

jazzy

### RMW implementation (if applicable):

_No response_

### RMW Configuration (if applicable):

_No response_

### Client library (if applicable):

_No response_

### 'ros2 doctor --report' output

<details><summary>ros2 doctor --report</summary>

```console
<COPY OUTPUT HERE>
```
</details>

### Steps to reproduce issue

The bug has been discovered while trying to include a C++ node with an exact time synchronizer policy in the [deterministic execution framework (ros2_def)](https://github.com/uulm-mrm/ros2_def). In short, the framework replicates the synchronizer policy in python in order to determine after which input an output of the node has to be expected. 

Reproducing the issue should however be possible with a simpler setup: 
Take two input streams and cause a reordering of samples. More specifically, having the two tuples (x1, x2) at timestamp t1 and (y1, y2) at timestamp t2 (with t1 < t2), the filter should for instance receive the inputs in the following order: x1, y1, y2 (causing a signal), x2. 
Depending on the chosen implementation the receival of x2 causes another callback since x1 is kept.

### Expected behavior

Given the same sequence of message inputs, the C++ and Python implementation should lead to the same sequence of synchronized outputs.

### Actual behavior

The C++ implementation drops all messages which are older than the latest signal time, while the python implementation retains them.

### Additional information

I discovered this already a while ago but it seems like I never opened an issue. By chance, I came across a similar issue in another repository (https://github.com/ros2/geometry2/issues/818), reminding me of this.
While skimming the updated code it seems like the implementation didn't change that much. Hence, I am still confident that the issue persists. However, I haven't explicitly retested it with the current rolling branch.
In case this issue has already been fixed in the past please close this issue.

More details:
The [C++ implementation](https://github.com/ros2/message_filters/blob/rolling/include/message_filters/sync_policies/exact_time.hpp#L153) erases all inputs older than the last signal time.

In contrast, the [python implementation](https://github.com/ros2/message_filters/blob/77c31e69acbb21e221fc521c1b652bf0c4d1cb45/src/message_filters/__init__.py#L233) keeps these older samples, thus, potentially causing another signalling once the missing message of the older tuple is received.

Out of curiosity:
What's the motivation behind the two independent implementations (in contrast to Python bindings)? As a side note: the two implementations do not even follow the same naming pattern for the policies.

For my use case I fixed it using the [following commit](https://github.com/authaldo/message_filters/tree/feat_more_cpp_python_consistency). If that is a suitable fix, I am happy to rebase it and contribute a PR.

EsipovPA:
## Description

In `Python3` implementation, messages with timestamps, older then the last that caused signalling, are removed and do not cause signalling in the future. As in `C++` implementation.

See the issue #200 

Fixes #200 

Yes.
Makes `Python3` behavior of the `TymeSynchronizer` filter consistent with `C++` behavior.

No, I did not

### Additional Information

@authaldo can you take a look at this one?
As an author of the original fix.

### DIFF
--- a/test/directed.py
+++ b/test/directed.py
@@ -41,6 +41,7 @@
 #
 #    wide.registerCallback(boost::bind(&PersonDataRecorder::wideCB, this, _1, _2, _3, _4));
 
+import functools
 import random
 import unittest
 
@@ -66,6 +67,7 @@ class MockFilter(SimpleFilter):
 
 class TestDirected(unittest.TestCase):
 
+    # TODO: Replace global variable with local
     def cb_collector_2msg(self, msg1, msg2):
         self.collector.append((msg1, msg2))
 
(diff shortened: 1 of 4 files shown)


test/directed.py
```diff
@@ -66,6 +67,7 @@ class MockFilter(SimpleFilter):
 
 class TestDirected(unittest.TestCase):
 
+    # TODO: Replace global variable with local
```
EsipovPA: Resolve this TODO

test/directed.py
```diff
@@ -105,8 +107,88 @@ def test_synchronizer(self):
...
+        filter_1.signalMessage(y1)
+        assert len(collector) == 2
+        assert collector[0] == (x0, x1)
+        assert collector[1] == (y0, y1)
+
 
 if __name__ == '__main__':
     suite = unittest.TestSuite()
     suite.addTest(TestDirected('test_synchronizer'))
+    suite.addTest(TestDirected('test_time_synchronizer_old_msgs_drop'))
```
EsipovPA: Add tests with their valid names here

EsipovPA: Done.

test/directed.py
EsipovPA: Done

ahcorde (approved):

ahcorde:
Pulls: ros2/message_filters#238
Gist: https://gist.githubusercontent.com/ahcorde/88b03f83de2070d5d156e5bec3ff252f/raw/b03fff1b8dbd98b0e104f6a0ba166484c7f5c4cf/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/17837
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/26704/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/20316/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/6503/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26291/)

authaldo:
Thanks @EsipovPA for fixing this!

ahcorde:
https://github.com/Mergifyio backport kilted jazzy humble

PR #256, opened 2026-03-06
Improvements and more test coverage

ahcorde:
## Description

Replace linear search with `std::lower_bound` and `std::upper_bound`

make local copies instead of holding the mutex for a long time

Improved test coverage

GitHub Copilot Claude Sonnet 4.6

### DIFF
--- a/include/message_filters/cache.hpp
+++ b/include/message_filters/cache.hpp
@@ -29,6 +29,7 @@
 #ifndef MESSAGE_FILTERS__CACHE_HPP_
 #define MESSAGE_FILTERS__CACHE_HPP_
 
+#include <algorithm>
 #include <cstddef>
 #include <deque>
 #include <functional>
@@ -153,33 +154,23 @@ public:
    */
   void add(const EventType & evt)
   {
-    namespace mt = message_filters::message_traits;
-
-    // printf("  Cache Size: %u\n", cache_.size());
     {
       std::lock_guard<std::mutex> lock(cache_lock_);
 
-      // Keep popping off old data until we have space for a new msg
-      // The front of the deque has the oldest elem, so we can get rid of it
+      // Keep popping off old data until we have space for a new msg.
+      // The front of the deque has the oldest elem.
       while (cache_.size() >= cache_size_) {
         cache_.pop_front();
       }
 
-      // No longer naively pushing msgs to back. Want to make sure they're sorted correctly
-      // cache_.push_back(msg);
-      // Add the newest message to the back of the deque
-
-      typename std::deque<EventType>::reverse_iterator rev_it = cache_.rbegin();
-
-      // Keep walking backwards along deque until we hit the beginning,
-      // or until we find a timestamp that's smaller than (or equal to) msg's timestamp
-      rclcpp::Time evt_stamp = getMessageTime(evt);
-      while (rev_it != cache_.rend() && getMessageTime(*rev_it) > evt_stamp) {
-        rev_it++;
-      }
-
-      // Add msg to the cache
-      cache_.insert(rev_it.base(), evt);
+      // Insert in sorted position using binary search (deque is kept time-ordered).
+      const rclcpp::Time evt_stamp = getMessageTime(evt);
+      auto it = std::lower_bound(
+        cache_.begin(), cache_.end(), evt_stamp,
+        [this](const EventType & e, const rclcpp::Time & t) {
+          return getMessageTime(e) < t;
+        });
+      cache_.insert(it, evt);
     }
 
     this->signalMessage(evt);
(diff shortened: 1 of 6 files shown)


ahcorde:
FYI @asymingt

include/message_filters/cache.hpp
```diff
@@ -153,33 +154,23 @@ class Cache : public SimpleFilter<M>
...
-      // Add msg to the cache
-      cache_.insert(rev_it.base(), evt);
+      // Insert in sorted position using binary search (deque is kept time-ordered).
+      const rclcpp::Time evt_stamp = getMessageTime(evt);
+      auto it = std::lower_bound(
+        cache_.begin(), cache_.end(), evt_stamp,
+        [this](const EventType & e, const rclcpp::Time & t) {
+          return getMessageTime(e) < t;
+        });
+      cache_.insert(it, evt);
```
asymingt, line 173: Why is `cache_` not a `std::map<rclcpp::Time, EventType>` if we care about log(N) insertion complexity? We could sill perform housekeeping on insertion to avoid unbounded size growth.

ahcorde: It only stores one value per key . Duplicate timestamps (two messages at the exact same rclcpp::Time) would silently overwrite each other. You'd need std::multimap<rclcpp::Time, EventType>.

If messages arrive mostly in-order (common for sensor streams), std::lower_bound on the deque resolves near the back quickly, and no shifting is needed, degenerating to amortized O(1) in the best case.

test/time_sequencer_unittest.cpp
```diff
@@ -149,6 +169,190 @@ TEST(TimeSequencer, eventInEventOut)
...
+  executor.add_node(node);
+
+  const rclcpp::Duration delay(0, 250000000);   // 250 ms
+  const rclcpp::Duration rate(0, 10000000);     // 10 ms
+  message_filters::TimeSequencer<Msg> seq(delay, rate, 10, node);
+
+  Helper h;
+  seq.registerCallback(std::bind(&Helper::cb, &h, std::placeholders::_1));
+
+  rclcpp::Time base = rclcpp::Clock(RCL_ROS_TIME).now();
```
asymingt, line 186: Is there any way we can avoid time-varying values as inputs to test fixtures, as it can lead to nondeterminism in your test infrastructure (stdout differences in the easy case, failures  in the hard case. If there's an easy way to make this constant, I'd suggest doing it.

test/time_sequencer_unittest.cpp
```diff
@@ -149,6 +169,190 @@ TEST(TimeSequencer, eventInEventOut)
...
+  for (int i = 0; i < n; ++i) {
+    auto msg = std::make_shared<Msg>();
+    // All timestamps are in the past relative to base, well beyond the delay
+    msg->header.stamp = base - rclcpp::Duration(1, 0) + rclcpp::Duration(0,
+      static_cast<uint32_t>(i) * 10000000u);
+    msg->data = i;
+    seq.add(msg);
+  }
+
+  ASSERT_TRUE(spinUntil(executor, [&h, n]() {return h.count_ >= n;}));
```
asymingt, line 330: This is a little bit of a nit, but If these are all meant to be dispatched in a single tick, it should be spinOnce, right?

include/message_filters/signal1.hpp
```diff
@@ -103,12 +103,13 @@ class Signal1
 
   void call(const MessageEvent<M const> & event)
   {
-    std::lock_guard<std::mutex> lock(mutex_);
-    bool nonconst_force_copy = callbacks_.size() > 1;
-    typename V_CallbackHelper1::iterator it = callbacks_.begin();
-    typename V_CallbackHelper1::iterator end = callbacks_.end();
-    for (; it != end; ++it) {
-      const CallbackHelper1Ptr & helper = *it;
+    V_CallbackHelper1 local_callbacks;
```
asymingt, line 106: The behavior has changed here to unlock before calling the nonconst callback copies. Looks like they are `std::shared_ptr` under the hood, so assume that this is OK. I just want to flag it in case it somehow is not.

include/message_filters/cache.hpp
asymingt: > It only stores one value per key . Duplicate timestamps (two messages at the exact same rclcpp::Time) would silently overwrite each other. You'd need std::multimap<rclcpp::Time, EventType>.
> 

Ah yes, whoops, that is what I meant!

> If messages arrive mostly in-order (common for sensor streams), std::lower_bound on the deque resolves near the back quickly, and no shifting is needed, degenerating to amortized O(1) in the best case.

I might be wrong, but every time you call `std::lower_bound` on a `deque` it's `log(N)`. and you're doing this on every insertion. The `std::deque` preserves insertion order, but our usage gives me the impression that we actually want value-ordering (by timestamp). So a lot of work is spent doing binary search for placement, which comes for free with a `std::multimap`.

Calling `rbegin()` and `begin()` on a multimap is O(1). Popping either off should also be low complexity because the internal structure is a red-black tree. And, you'd also be able to easily extract ranges with: https://en.cppreference.com/w/cpp/container/multimap/equal_range.html

The only argument against using a multimap is memory footprint, but for typical queue sizes of less than 1M items I doubt you'll notice it.

My sense is that you'll halve the lines of code by using a multimap here. Regardless of what you choose, it would be interesting to benchmark this.

ahcorde: @asymingt is it fine ? can you approve the PR ?

asymingt (approved):
I think this is an improvement upon what was already there, so LGTM. But I do think we should backlog an  issue to benchmark a multimap in place of the deque. Mostly because I think it will reduce the lines of code with little to no effect on runtime performance.

ahcorde:
Pulls: ros2/message_filters#256
Gist: https://gist.githubusercontent.com/ahcorde/d5ccc9b480f5cce2bb9823c1168768d7/raw/01f96a5d0e8fe1122c5542b83bc0219c73496577/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/18638
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/28257/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21182/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/8087/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/27174/)

Issue #130, opened 2024-06-08
Extend Tutorials Section


CursedRock17:
As per the title we could keep expanding the [tutorials section](https://github.com/ros2/message_filters/tree/rolling/doc/Tutorials) of our `message_filters` documentation because you can never have enough documentation. Specifically https://github.com/ros2/message_filters/pull/119#discussion_r1604565909 specifies for us to add an [ApproximateEpsilonTime](https://github.com/ros2/message_filters/blob/rolling/include/message_filters/sync_policies/approximate_epsilon_time.h) tutorials. That being said the following policies and objects could technically have their own tutorials:

- [ ] [ApproximateEpsilonTime](https://github.com/ros2/message_filters/blob/rolling/include/message_filters/sync_policies/approximate_epsilon_time.h) (C++)
- [ ] [LatestTime](https://github.com/ros2/message_filters/blob/ac26507686ea8533504bb917128659951029759d/include/message_filters/sync_policies/latest_time.h) (C++)
- [ ] [Chain](https://github.com/ros2/message_filters/blob/ac26507686ea8533504bb917128659951029759d/include/message_filters/chain.h#L104) (C++)
- [x] [Cache](https://github.com/ros2/message_filters/blob/ac26507686ea8533504bb917128659951029759d/include/message_filters/cache.h#L62) (C++)
- [x] [Cache](https://github.com/ros2/message_filters/blob/ac26507686ea8533504bb917128659951029759d/src/message_filters/__init__.py#L93) (Python)

Possibly:

- [ ] [Custom Simple Filter](https://github.com/ros2/message_filters/blob/ac26507686ea8533504bb917128659951029759d/include/message_filters/simple_filter.h#L51) (C++ - Intermediate)
- [ ] [Custom Simple Filter](https://github.com/ros2/message_filters/blob/ac26507686ea8533504bb917128659951029759d/src/message_filters/__init__.py#L46) (Python - Intermediate)

I should be able to get to most of these relatively fast, but they should be in one location ready for tracking. Additionally since most of tfhe documentation is hosted on `message_filters` this shouldn't need to go to `ros2_documentation`

EsipovPA:
Is someone working on this one? If nobody is, is it ok, If I try to add at least some of them?

Looking for some issues to work on in some spare time. 🙂 

upd

I've started with the `Cache` filter for `Python`, as I've worked with lately. If anybody is OK with it.

CursedRock17:
[...]

PR #261, opened 2026-03-10
130 add approximate epsilon time tutorial

EsipovPA:
## Description

1) Add C++ tutorial for `ApproximateEpsilonTime` Sync policy
2) Label CMake code blocks with the right language markings in the `Approximate-Synchronizer` tutorial

Relates to #130 

Does not change the behavior.
Affects UX.

No I did not.

### DIFF
--- a/doc/Tutorials/Approximate-Epsilon-Tyme-Synchronizer-Cpp.rst
+++ b/doc/Tutorials/Approximate-Epsilon-Tyme-Synchronizer-Cpp.rst
@@ -1,0 +1,442 @@
+Approximate Epsilon Time Synchronizer (C++):
+---------------------------------------
+
+Prerequisites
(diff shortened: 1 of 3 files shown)


EsipovPA:
Is it possible not to squash commits when merging this PR? I may add a PR id to commit messages if it is required.

The motivation is to keep the changes in the ApproximateTimeSync tutorialwith CMake marking separate from the rest of PR. It seems like an overkill to open a separate PR for this kind of change, but it does not really fit into the major commit in this PR. I may move it to a separate commit though, but if it is possible, it would be great to be able to merge them as soon as possible.

ahcorde (approved):

ahcorde:
https://github.com/Mergifyio backport kilted jazzy humble

PR #283, opened 2026-04-24
feat(python): add python implementation of InputAligner 

### LINKED ISSUE #282: Add a Python implementation of InputAligner
## Summary
The package currently provides a C++ `InputAligner` implementation in `include/message_filters/input_aligner.hpp`, but there is no equivalent Python API in `message_filters/__init__.py`.

It would be useful to add a Python implementation with behavior aligned to the existing C++ version, including parity tests for the core dispatch / timeout / drop semantics.

## Motivation
Python users of `message_filters` currently do not have access to `InputAligner`, while several other filters already provide both C++ and Python interfaces. Adding Python support would make the API surface more consistent across languages.

## Proposed scope
- add `message_filters.input_aligner.InputAligner` in Python
- export it from the Python package
- add parity-style Python tests covering ordering, inactive inputs, timeout behavior, dropped messages, timer dispatch, and queue status

## Related work
A downstream rolling PR already exists here:
- UniflexAI/message_filters#5

That PR cherry-picks the humble implementation work onto rolling and can be used as a reference when discussing an upstreamable version.

dvorak0:
## Description

Fixes https://github.com/ros2/message_filters/issues/282

Codex

## Summary
- add a Python `InputAligner` implementation on top of the `rolling` branch
- export it from the Python package
- add Python parity tests matching the C++ `test_input_aligner.cpp` coverage

## Details
This PR targets `ros2/message_filters:rolling` from `UniflexAI/message_filters:rolling`.
The Python implementation follows the C++ InputAligner semantics referenced from the upstream `sa-input_aligner` work.

## Test coverage
The Python test file now covers 10 InputAligner cases, including the original C++ parity coverage plus Python-specific regression coverage for recent fixes:
- init
- dispatch_inputs_in_order
- dispatch_inputs_with_duplicate_timestamps
- reconnect_input_disconnects_old_callbacks
- ignores_inactive_inputs
- input_timeout
- drops_msgs
- dispatch_by_timer
- no_period_information
- get_queue_status

## Validation
Validated on a Rolling environment with all 10 Python InputAligner tests passing.

### DIFF
--- a/src/message_filters/__init__.py
+++ b/src/message_filters/__init__.py
@@ -50,6 +50,8 @@ from rclpy.subscription_content_filter_options import ContentFilterOptions
 from rclpy.time import Time
 from rclpy.type_support import MsgT
 
+from .input_aligner import InputAligner, QueueStatus
+
 
 class SimpleFilter(object):
 
--- a/src/message_filters/__init__.py
+++ b/src/message_filters/__init__.py
@@ -34,7 +34,7 @@ from dataclasses import dataclass
 from functools import reduce
 import itertools
 import threading
-from typing import Optional, Type, TypeVar, Union
+from typing import Optional, Type, Union
 
 from builtin_interfaces.msg import Time as TimeMsg
 import rclpy
@@ -46,15 +46,9 @@ from rclpy.logging import LoggingSeverity
 from rclpy.node import Node
 from rclpy.qos import QoSProfile
 from rclpy.qos_overriding_options import QoSOverridingOptions
-try:
-    from rclpy.subscription_content_filter_options import ContentFilterOptions
-except ImportError:
-    ContentFilterOptions = None
+from rclpy.subscription_content_filter_options import ContentFilterOptions
 from rclpy.time import Time
-try:
-    from rclpy.type_support import MsgT
-except ImportError:
-    MsgT = TypeVar("MsgT")
+from rclpy.type_support import MsgT
 
 from .simple_filter import SimpleFilter
 from .input_aligner import InputAligner
--- a/src/message_filters/__init__.py
+++ b/src/message_filters/__init__.py
@@ -46,15 +46,9 @@ from rclpy.logging import LoggingSeverity
 from rclpy.node import Node
 from rclpy.qos import QoSProfile
 from rclpy.qos_overriding_options import QoSOverridingOptions
-try:
-    from rclpy.subscription_content_filter_options import ContentFilterOptions
-except ImportError:
-    ContentFilterOptions = None
+from rclpy.subscription_content_filter_options import ContentFilterOptions
 from rclpy.time import Time
-try:
-    from rclpy.type_support import MsgT
-except ImportError:
-    MsgT = TypeVar("MsgT")
+from rclpy.type_support import MsgT
 
 
 class SimpleFilter(object):
@@ -144,7 +138,7 @@ class Subscriber(SimpleFilter):
         return self.sub.__getattribute__(key)
 
 
-from message_filters.input_aligner import InputAligner, QueueStatus
+from message_filters.input_aligner import InputAligner
 
 
 class Cache(SimpleFilter):
@@ -588,4 +582,3 @@ class TimeSequencer(SimpleFilter):
         """Clean up the TimeSequencer."""
         self.update_timer.cancel()
         self.node.destroy_timer(self.update_timer)
-
--- a/src/message_filters/__init__.py
+++ b/src/message_filters/__init__.py
@@ -50,6 +50,7 @@ from rclpy.subscription_content_filter_options import ContentFilterOptions
 from rclpy.time import Time
 from rclpy.type_support import MsgT
 
+from message_filters.input_aligner import InputAligner, QueueStatus
 
 
 class SimpleFilter(object):
@@ -581,4 +582,3 @@ class TimeSequencer(SimpleFilter):
         self.update_timer.cancel()
         self.node.destroy_timer(self.update_timer)
 
-from message_filters.input_aligner import InputAligner, QueueStatus
--- a/src/message_filters/__init__.py
+++ b/src/message_filters/__init__.py
@@ -72,6 +72,9 @@ class SimpleFilter(object):
         for (cb, args) in self.callbacks.values():
             cb(*(msg + args))
 
+    def unregisterCallback(self, conn):
+        self.callbacks.pop(conn, None)
+
 
 class Subscriber(SimpleFilter):
     """
--- a/src/message_filters/__init__.py
+++ b/src/message_filters/__init__.py
@@ -34,7 +34,7 @@ from dataclasses import dataclass
 from functools import reduce
 import itertools
 import threading
-from typing import Optional, Type, TypeVar, Union
+from typing import Optional, Type, Union
 
 from builtin_interfaces.msg import Time as TimeMsg
 import rclpy
--- a/src/message_filters/__init__.py
+++ b/src/message_filters/__init__.py
@@ -51,7 +51,6 @@ from rclpy.time import Time
 from rclpy.type_support import MsgT
 
 from .simple_filter import SimpleFilter
-from .input_aligner import InputAligner
 
 
 class Subscriber(SimpleFilter):
--- a/src/message_filters/__init__.py
+++ b/src/message_filters/__init__.py
@@ -34,7 +34,7 @@ from dataclasses import dataclass
 from functools import reduce
 import itertools
 import threading
-from typing import Optional, Type, Union
+from typing import Optional, Type, TypeVar, Union
 
 from builtin_interfaces.msg import Time as TimeMsg
 import rclpy
@@ -46,11 +46,15 @@ from rclpy.logging import LoggingSeverity
 from rclpy.node import Node
 from rclpy.qos import QoSProfile
 from rclpy.qos_overriding_options import QoSOverridingOptions
-from rclpy.subscription_content_filter_options import ContentFilterOptions
+try:
+    from rclpy.subscription_content_filter_options import ContentFilterOptions
+except ImportError:
+    ContentFilterOptions = None
 from rclpy.time import Time
-from rclpy.type_support import MsgT
-
-from message_filters.input_aligner import InputAligner, QueueStatus
+try:
+    from rclpy.type_support import MsgT
+except ImportError:
+    MsgT = TypeVar("MsgT")
 
 
 class SimpleFilter(object):
@@ -138,6 +142,9 @@ class Subscriber(SimpleFilter):
     def __getattr__(self, key):
         """Serve same API as rospy.Subscriber."""
         return self.sub.__getattribute__(key)
+
+
+from message_filters.input_aligner import InputAligner, QueueStatus
 
 
 class Cache(SimpleFilter):
--- a/src/message_filters/__init__.py
+++ b/src/message_filters/__init__.py
@@ -34,7 +34,7 @@ from dataclasses import dataclass
 from functools import reduce
 import itertools
 import threading
-from typing import Optional, Type, Union
+from typing import Optional, Type, TypeVar, Union
 
 from builtin_interfaces.msg import Time as TimeMsg
 import rclpy
@@ -46,34 +46,18 @@ from rclpy.logging import LoggingSeverity
 from rclpy.node import Node
 from rclpy.qos import QoSProfile
 from rclpy.qos_overriding_options import QoSOverridingOptions
-from rclpy.subscription_content_filter_options import ContentFilterOptions
+try:
+    from rclpy.subscription_content_filter_options import ContentFilterOptions
+except ImportError:
+    ContentFilterOptions = None
 from rclpy.time import Time
-from rclpy.type_support import MsgT
+try:
+    from rclpy.type_support import MsgT
+except ImportError:
+    MsgT = TypeVar("MsgT")
 
-
-class SimpleFilter(object):
-
-    def __init__(self):
-        self.callbacks = {}
-
-    def registerCallback(self, cb, *args):
-        """
-        Register a callback `cb` to be called when this filter has output.
-
-        The filter calls the function ``cb`` with a filter-dependent.
-
-        list of arguments,followed by the call-supplied arguments ``args.``.
-        """
-        conn = len(self.callbacks)
-        self.callbacks[conn] = (cb, args)
-        return conn
-
-    def signalMessage(self, *msg):
-        for (cb, args) in self.callbacks.values():
-            cb(*(msg + args))
-
-    def unregisterCallback(self, conn):
-        self.callbacks.pop(conn, None)
+from .simple_filter import SimpleFilter
+from .input_aligner import InputAligner
 
 
 class Subscriber(SimpleFilter):
@@ -139,9 +123,6 @@ class Subscriber(SimpleFilter):
     def __getattr__(self, key):
         """Serve same API as rospy.Subscriber."""
         return self.sub.__getattribute__(key)
-
-
-from message_filters.input_aligner import InputAligner
 
 
 class Cache(SimpleFilter):
--- a/src/message_filters/__init__.py
+++ b/src/message_filters/__init__.py
@@ -50,7 +50,6 @@ from rclpy.subscription_content_filter_options import ContentFilterOptions
 from rclpy.time import Time
 from rclpy.type_support import MsgT
 
-from .input_aligner import InputAligner, QueueStatus
 
 
 class SimpleFilter(object):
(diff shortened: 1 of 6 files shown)


dvorak0:
@EsipovPA Here is the rolling version. Would you help again? Thanks.

src/message_filters/input_aligner.py
```diff
@@ -0,0 +1,168 @@
+from dataclasses import dataclass
```
ahcorde, line 31: Missing License header

ahcorde (changes requested):
1. Tie-break crash — input_aligner.py:108 puts (timestamp, msg) into a PriorityQueue. Equal timestamps fall through to comparing messages → TypeError. Add a sequence counter: (ts, seq,    
  msg). The C++ version (a std::multiset keyed on timestamp) sidesteps this.
  2. disconnectAll() doesn't disconnect — input_aligner.py:81. It clears only the local list; upstream SimpleFilter callbacks stay live. Re-calling connectInput with fewer inputs →          
  IndexError on the next upstream message. SimpleFilter (__init__.py:54) has no unregister API — that needs to be added.                                                                      
  3. Race in connectInput — input_aligner.py:72-79 replaces event_queues/signals without the lock while add can still fire.

dvorak0:
@ahcorde I've fixed the issues, and added two more tests coordinating to issue 1 & 2.

src/message_filters/__init__.py
```diff
@@ -138,6 +141,9 @@ def __getattr__(self, key):
         return self.sub.__getattribute__(key)
 
 
+from message_filters.input_aligner import InputAligner
```
EsipovPA: If you are aiming to add this class to the `__init__.py`, maybe move the `SimpleFiler` class to a separate file and import it here and in `input_aligner.py`? This way you'll be able to move this import to the top of this file with the rest of the `import` section.

src/message_filters/input_aligner.py
```diff
@@ -0,0 +1,204 @@
...
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+"""Input aligner for synchronizing messages from multiple sources based on their timestamps."""
+
+from dataclasses import dataclass
+from queue import PriorityQueue as Queue
```
EsipovPA: This alias uses the already taken name from the standard library. It'll work, but it messes up the naming. Maybe use the original class name? `PriorityQueue` does not look that long, to require such aliases. 

Just a proposal. Not an absolute requirement.

Also, maybe move these `from` imports below library imports that are `import threading` and `import typing as tp`?

src/message_filters/input_aligner.py
```diff
@@ -0,0 +1,204 @@
...
+
+@dataclass
+class QueueStatus:
+    active: bool
+    queue_size: int
+    msgs_processed: int
+    msgs_dropped: int
+
+
+class _Signal:
```
EsipovPA: Maybe make it an internal class of the `InputAligner` as it is used only in that class and is not intended to be used outside of it.

src/message_filters/__init__.py
dvorak0: I like the idea of having simple_filter.py.

src/message_filters/input_aligner.py
dvorak0: will do.

src/message_filters/input_aligner.py
```diff
@@ -0,0 +1,204 @@
...
+    def __init__(self):
+        self.callbacks = {}
+
+    def registerCallback(self, cb, *args):
```
EsipovPA:
```suggestion
    def registerCallback(
        self,
        callback: tp.Callable[
            [
                expected,
                argument,
                types
            ],
            expected_return_type
        ],
        *args
    ) -> int:
```
As the `typing` is already imported, why not use it to the full extend? Is the return type an `int`?

src/message_filters/input_aligner.py
```diff
@@ -0,0 +1,204 @@
...
+class QueueStatus:
+    active: bool
+    queue_size: int
+    msgs_processed: int
+    msgs_dropped: int
+
+
+class _Signal:
+    def __init__(self):
+        self.callbacks = {}
```
EsipovPA:
```suggestion
        self.callbacks: dict[key_type, value_type] = {}
```
It seems to me that the `value_type` is something like `tp.Callable[[...], ...]`

src/message_filters/__init__.py
dvorak0: do we need to do this for other filters, like time sequencer?

src/message_filters/input_aligner.py
```diff
@@ -0,0 +1,204 @@
...
+
+    def registerCallback(self, cb, *args):
+        conn = len(self.callbacks)
+        self.callbacks[conn] = (cb, args)
+        return conn
+
+    def signalMessage(self, *msg):
+        for (cb, args) in self.callbacks.values():
```
EsipovPA:
```suggestion
        for (callback, args) in self.callbacks.values():
```
Makes it more readable IMHO.

src/message_filters/input_aligner.py
```diff
@@ -0,0 +1,204 @@
...
+
+
+def _ros_max_time() -> Time:
+    return Time(
+        nanoseconds=9223372036854775807,
+        clock_type=_ros_zero_time().clock_type,
+    )
+
+
+class _EventQueue:
```
EsipovPA: [Same here](https://github.com/ros2/message_filters/pull/283/changes#r3141774463)

src/message_filters/input_aligner.py
```diff
@@ -0,0 +1,204 @@
...
+
+    def first_timestamp(self) -> Time:
+        if not self.events.empty():
+            first_ts = self.events.queue[0][0]
+            self.next_ts = first_ts + self.period
+            self.active = True
+            return first_ts
+        if self.active:
+            return self.next_ts
+        return _ros_max_time()
```
EsipovPA:
```suggestion
        return self.next_ts
```
It seems like the `self.active` never gets `False` after [this](https://github.com/ros2/message_filters/pull/283/changes#diff-4be9a867ec5f402f60f382a051e6eb3de96c48da415981d8196138cdb2ce858bR89).

Also, the initial value of `self.next_ts` is already `_ros_max_time`. So, this doesn't have any difference than  this proposal. 

Also, if not the first option, I would suggest something like this

```suggestion
        elif self.active:
            return self.next_ts
        return _ros_max_time()
```

But this is just a taste preference. Doesn't make any difference in how the code is executed. So it is totally up to you.

src/message_filters/input_aligner.py
```diff
@@ -0,0 +1,204 @@
...
+
+    def pop_first(self) -> None:
+        self.events.get_nowait()
+        self.msgs_processed += 1
```
EsipovPA:
```suggestion
    def pop_first(self) -> EventType | None:
        if not self.events.queue.empty():
            self.msgs_processed += 1
            return self.events.pop()
        return None
```

A `pop` method, usually is expected to return a value and remove it from a container.

src/message_filters/input_aligner.py
```diff
@@ -0,0 +1,204 @@
...
+
+    def get_status(self) -> QueueStatus:
+        return QueueStatus(self.active, self.events.qsize(), self.msgs_processed, self.msgs_dropped)
+
+
+class InputAligner(SimpleFilter):
+    def __init__(
+        self,
+        timeout: Duration,
+        *filters: SimpleFilter
```
EsipovPA:
```suggestion
        filters: list[SimpleFilter],
```
? At least the type hint is wrong.

src/message_filters/input_aligner.py
```diff
@@ -0,0 +1,204 @@
...
+        self.event_queues: list[_EventQueue] = []
+        self.input_connections: list[tuple[SimpleFilter, int]] = []
+        self.signals: list[_Signal] = []
+        self.dispatch_timer: tp.Any = None
+        if filters:
+            self.connectInput(*filters)
+
+    def connectInput(
+        self,
+        *filters: SimpleFilter
```
EsipovPA:
```suggestion
        filters: list[SimpleFilter],
```

src/message_filters/input_aligner.py
```diff
@@ -0,0 +1,204 @@
...
+
+    def disconnectAll(self) -> None:
+        for input_filter, conn in self.input_connections:
+            input_filter.unregisterCallback(conn)
+        self.input_connections = []
+
+    def registerCallback(
+        self,
+        index: int,
+        callback: tp.Callable[..., tp.Any],
```
EsipovPA: IMHO, this type hint is not really informative. It is the same as just
```suggestion
        callback: tp.Callable,
```
But it does not privide info about the callback signature. Myabe extend this type hint?

src/message_filters/input_aligner.py
dvorak0: Do you recommend to switch `aligner.connectInput(f0, f1, f2)` to `aligner.connectInput([f0, f1, f2])`? If we accept filters as `aligner.connectInput(f0, f1, f2)`, its type hint is `*filters: SimpleFilter`.

src/message_filters/input_aligner.py
dvorak0: I'm thinking ` *filters: SimpleFilter` means: varargs with the same type. https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html#functions 

```
# This says each positional arg and each keyword arg is a "str"
def call(self, *args: str, **kwargs: str) -> str:
    reveal_type(args)  # Revealed type is "tuple[str, ...]"
    reveal_type(kwargs)  # Revealed type is "dict[str, str]"
    request = make_request(*args, **kwargs)
    return self.do_api_query(request)
```

src/message_filters/input_aligner.py
dvorak0: I agreed the info is too loose. How can I do better when they are callbacks, which means I don't know about it much. What do you recommend to do.

src/message_filters/input_aligner.py
EsipovPA: Yes, it does. But they function as a list anyway. The con of this approach is that you are unable to pass `filters` as a keyword argument. Also it makes using and extending this signature a bit complicated.  IMHO, the best case for *args, **kwargs is when you pass them to some other function (classmethod, whatever) that may have varying signature itself  depending on a specific case. What I propose is to make uaer interface precise and completely defined. It makes using a class easier.

src/message_filters/input_aligner.py
EsipovPA: The idea is to make possible the following call

```
aligner.connectInput(
    filters=[filters_list]
)
```
The current approach does not permit using `filters` as a keyword argument. Also, arguments that may be added after `filters` will be usable only as kwargs. Not as positional arguments.

src/message_filters/input_aligner.py
EsipovPA: I'll need to take a look into the source code for that. I'll be back with an answer.

dvorak0: I think I've done all others changes as requested, expect for this one.

EsipovPA: Sorry, I've been avay from a computer for a few days. I'll try to return with the answer today (29.04.26)

src/message_filters/input_aligner.py
```diff
@@ -0,0 +1,206 @@
+# Copyright 2009, Willow Garage, Inc. All rights reserved.
```
ahcorde:
```suggestion
# Copyright 2026 Open Source Robotics Foundation, Inc.
```

src/message_filters/simple_filter.py
```diff
@@ -0,0 +1,52 @@
+# Copyright 2009, Willow Garage, Inc. All rights reserved.
```
ahcorde:
```suggestion
# Copyright 2026 Open Source Robotics Foundation, Inc.
```

test/test_input_aligner.py
```diff
@@ -0,0 +1,236 @@
+import time
```
ahcorde, line 30: Add License header

ahcorde (changes requested):
I added some suggestions here 

https://github.com/ros2/message_filters/tree/pr-283

can you can cherry-pick them

dvorak0:
> I added some suggestions here
> 
> https://github.com/ros2/message_filters/tree/pr-283
> 
> can you can cherry-pick them

@ahcorde Thanks. Just cherry-picked your commit. Tests passed.

src/message_filters/input_aligner.py
EsipovPA: Well, I was away from computer again. Here is the idea: If everything is ok, maybe change the typing description to ot's original form `callback: tp.Callable` and add a TODO that goes something like 

```
callback: tp.Callable,    #TODO: @EsipovPA Fix typing for callable
```

and close this thread? I'll fix typing when I'm back behind the ceyboard.

The idea is to roll this PR further and stop blocking it with this thread. If the codeowners are OK with it, I'm also OK. LFTM.

src/message_filters/delta_filter.py
```diff
@@ -31,7 +31,7 @@
 from functools import partial
 from typing import Any, Callable
 
-from message_filters import SimpleFilter
+from .simple_filter import SimpleFilter
```
EsipovPA:
```suggestion
from message_filters.simple_filter import SimpleFilter
```

Same with all the other relative imports. IMHO, absolute imports > relative imports

src/message_filters/input_aligner.py
```diff
@@ -0,0 +1,214 @@
...
+            self.msgs_dropped += 1
+
+        def set_period(self, period: Duration) -> None:
+            self.period = period
+
+        def set_active(self, active: bool) -> None:
+            self.active = active
+
+        def get_status(self) -> QueueStatus:
+            return QueueStatus(self.active, len(self.events), self.msgs_processed, self.msgs_dropped)
```
EsipovPA:
```suggestion
            return QueueStatus(
                active=self.active,
                queue_size=len(self.events),
                msgs_processed=self.msgs_processed,
                msgs_dropped=self.msgs_dropped
            )
```

EsipovPA (approved):
LFTM regarding my comments. I have nothing to add to what was commented before.

dvorak0:
@ahcorde Just refined the licence header. change relative imports to be absolute. and break the long line.

Thanks for your detailed reviews.

EsipovPA (approved):
Can't see any major issues. Overall LFTM. Let's see, how it will perform in the real world.

ahcorde (approved):

dvorak0:
@Mergifyio queue

dvorak0:
@ahcorde @EsipovPA Shall we merge this? Let me know if there is anything I need to do.

ahcorde:
Pulls: ros2/message_filters#283
Gist: https://gist.githubusercontent.com/ahcorde/8b3b4aeb51292741cdeb1c04a0060433/raw/3434868e63080542c571c117b936fe2ec952c30e/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/19104
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/28898/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21698/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/8761/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/27773/)

dvorak0:
@ahcorde I found CI failed due to lints, and fixed them. Could you re-trigger the pull/test. Thanks.

ahcorde (changes requested):
Do you mind to fix the linter issues?

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/28904/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21704/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/8767/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/27779/)

ahcorde (approved):

ahcorde:
https://github.com/Mergifyio backport lyrical kilted jazzy humble

PR #304, opened 2026-05-25
(#221) Tutorials: Add DeltaFilter C++ tutorial

### LINKED ISSUE #221: UnchangedFieldsFilter (WIP naming) Filter to drop messages that do not contain changed fields
### Description

- [X] Python implementation
- [x] Python documentation
- [x] C++ implementation
- [ ] C++ Documentation

*disclamer1*
The proposed class name is not the final one. Better naming required.

*disclamer2*
This is just an idea, that appeared to me today. I am not not sure if it should be implemented as a part of this library, or if it is just an overengineering of sorts for such a trivial problem. Wanted to discuss it.

**(Sort of a)Problem to solve**

In my work I oftentimes encounter a need to drop the messages, that do not contain changed data in a specific fields. Like the messages from automobile CAN-bus, that contain data about vehicle body. These messages are published with the frequency ~100Hz, but contain all the same data. It is only an example

**The idea**

The idea is to implement a filter, that drops messages, that do not contain changes in the specified fields, or all the fields, excluding headers.

**How it is solved now**

This is purely for demonstration.
Say there is a node with a subscription.

```
from rclpy_message_converter.message_converter import convert_ros_message_to_dictionary

class ExampleNode:
    def __init__(self, node, *args, **kwargs):
        self.node = node
        self.subscription = node.create_subscription(...., self.sub_callback)
        self.last_data: dict | None = None

    def sub_callback(self, message):
        # Suppose they may be compared this way
        # Suppose the `message` contains only data used further
        # This comparison may be much more complicated and take more operations
        # Gets more and more complicated with more messages and subscriptions
        new_data = convert_ros_message_to_dictionary(message)
        if new_data != self.last_data:
            # If fields changed, do some work here

```

**How it may look in my opinion**

For example, the python implementation may have the following interface

```
class UnchangedFieldsFilter(SimpleFilter):
    def __init__(
        self,
        other_filter,       # Supposedly a Subscriber filter
        include_fields_regex: list[str]=[],         # In this case, the empty list means all fields except a Header
        exclude_fields_regex: list[str]=[],        # Same thing, but for exclusion
    ):
        # The main implementation and all other stuff
```
So, with the use of this filter the previous code example may be converted to 
```
class ExampleNode:
    def __init__(self, node, *args, **kwargs):
        self.node = node
        self.subscriber_filter = Subscriber(# Initialize here #)
        self.unchanged_fields_filter = UnchangedFieldsFilter(
            self.subscriber_filter
            ['fields', 'expected', 'to', 'change'],
            ['fields', 'to', 'ignore', 'changes', 'in'],
        )
        self.unchanged_fields_filter.registerCallback(self.sub_callback)

    def sub_callback(self, message):
        # Do the work that requires updated data straight away
```
So for any subscription, added to the node, that needs only changed data, it will only require to add a `Subscriber` filter, an `UnchangedFieldsFilter` and a callback for the latter.

### Motivation

- Clearer code for nodes with multiple subscriptions that have the requirements, described in the **(Sort of a)Problem to solve** section of the description.
- Simpler code support for such nodes as a result.

### Design / Implementation Considerations

- Needs a better name, definetely.
- Interfaces for Python and C++ may be better. What is proposed is just a concept.
- Does someone see this as a useful thing in general? I have my own implementations for such scenarios, as most of other developers, using ROS2, I think. But it seems like there should be something that may appear in a `message_filters`, doing this king of stuff.

EsipovPA:
## Description

Added a tutorial for `DeltaFilter` for C++.

Fixes #221 
Final PR to close the issue.

No I did not.

### DIFF
--- a/doc/Tutorials/Delta-Filter-Cpp.rst
+++ b/doc/Tutorials/Delta-Filter-Cpp.rst
@@ -1,0 +1,530 @@
+DeltaFilter (C++):
+------------------
+
+Overview
+~~~~~~~~
+
+This tutorial demonstrates how to use the ``message_filters::DeltaFilter`` class in ROS 2 using C++.
+With the ``DeltaFilter`` given a stream of messages, the message is passed down
+to the next filter if any of the message fields, that may be acquired
+by ``field_getters`` have changed compared to the previously accepted message.
+
+To demonstrate the functionality of the ``DeltaFilter`` filter in a more clear manner, we are going to add a custom filter to this tutorial.
+This is going to be the ``CounterFilter`` that will be counting the number of messages passing through it.
+This filter class will be a successor to the ``SimpleFilter`` class, but this is a topic for another tutorial.
+
+For more information on this succession mechanism, please refer to the `SimpleFilter for C++ tutorial <https://docs.ros.org/en/rolling/p/message_filters/doc/Tutorials/SimpleFilter-Cpp.html>`_.
+
+Prerequisites
+~~~~~~~~~~~~~
+
+This tutorial assumes you have a working knowledge of ROS 2.
+
+If you have not done so already `create a workspace <https://docs.ros.org/en/rolling/Tutorials/Beginner-Client-Libraries/Creating-A-Workspace/Creating-A-Workspace.html>`_ and `create a package <https://docs.ros.org/en/rolling/Tutorials/Beginner-Client-Libraries/Creating-Your-First-ROS2-Package.html>`_
+
+1. Create a Basic Node with Includes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Let's assume, you've already created an empty ROS 2 package for C++.
+The next step is to create a new C++ file inside your package, e.g., ``delta_tutorial.cpp``, and write an example code:
+
+.. code-block:: C++
+
+  #include <chrono>
+  #include <functional>
+  #include <string>
+
+  #include <rclcpp/rclcpp.hpp>
+
+  #include <message_filters/delta_filter.hpp>
+  #include <message_filters/subscriber.hpp>
+
+  #include <std_msgs/msg/string.hpp>
+
+  using namespace std::chrono_literals;
+
+  const std::string TUTORIAL_TOPIC_NAME = "tutorial_topic";
+  const rclcpp::QoS QOS_PROFILE = rclcpp::QoS(10);
+
+  class CounterFilter : public message_filters::SimpleFilter<std_msgs::msg::String> {
+  public:
+    typedef message_filters::MessageEvent<std_msgs::msg::String const> EventType;
+
+    CounterFilter() : message_filters::SimpleFilter<std_msgs::msg::String>() {}
+
+    void add(const EventType & evt) {
+      ++counter_;
+      signalMessage(evt);
+    }
+
+    template<class F>
+    void connectInput(F & f) {
+      incoming_connection_.disconnect();
+      incoming_connection_ = f.registerCallback(
+        message_filters::SimpleFilter<std_msgs::msg::String>::EventCallback(
+          std::bind(
+            &CounterFilter::add,
+            this,
+            std::placeholders::_1
+          )
+        )
+      );
+    }
+
+    size_t getCounterValue() {
+      return counter_;
+    }
+  private:
+    size_t counter_ = 0;
+
+    message_filters::Connection incoming_connection_;
+  };
+
+  class DeltaNode : public rclcpp::Node {
+  public:
+    typedef std::shared_ptr<std_msgs::msg::String const> MConstPtr;
+
+    DeltaNode():
+      Node("DeltaNode"),
+      subscriber_filter_(),
+      delta_filter_(
+        subscriber_filter_,
+        {[] (const MConstPtr& msg) {return msg->data;}}
+      ),
+      counter_filter_()
+    {
+      subscriber_filter_.subscribe(this, TUTORIAL_TOPIC_NAME, QOS_PROFILE);
+
+      publisher_ = create_publisher<std_msgs::msg::String>(TUTORIAL_TOPIC_NAME, QOS_PROFILE);
+
+      counter_filter_.connectInput(delta_filter_);
+
+      publisher_timer_ = this->create_wall_timer(
+        1s,
+        std::bind(&DeltaNode::publisher_timer_callback, this)
+      );
+
+      query_timer_ = this->create_wall_timer(
+        1s,
+        std::bind(&DeltaNode::query_timer_callback, this)
+      );
+    }
+
+    void query_timer_callback() {
+      RCLCPP_INFO(
+        get_logger(),
+        "Messages passed down the chain count: %zu",
+        counter_filter_.getCounterValue()
+      );
+    }
+
+    void publisher_timer_callback() {
+      auto message = std_msgs::msg::String();
+      if (pub_count_ % 2 == 0) {
+        message.data = "pub_count = " + std::to_string(pub_count_);
+      } else {
+        message.data = "pub_count = " + std::to_string(pub_count_ - 1);
+      }
+      publisher_->publish(message);
+      ++pub_count_;
+    }
+
+  private:
+    message_filters::Subscriber<std_msgs::msg::String> subscriber_filter_;
+
+    message_filters::DeltaFilter<std_msgs::msg::String> delta_filter_;
+    CounterFilter counter_filter_;
+
+    rclcpp::Publisher<std_msgs::msg::String>::SharedPtr publisher_;
+
+    rclcpp::TimerBase::SharedPtr publisher_timer_;
+    rclcpp::TimerBase::SharedPtr query_timer_;
+
+    int pub_count_ { 0 };
+  };
+
+  int main([[maybe_unused]] int argc, [[maybe_unused]] char ** argv)
+  {
+    rclcpp::init(argc, argv);
+
+    auto delta_node = std::make_shared<DeltaNode>();
+
+    rclcpp::spin(delta_node);
+    rclcpp::shutdown();
+
+    return 0;
+  }
+
+1.1 Examine the code
+~~~~~~~~~~~~~~~~~~~~
+
+Now, let's break down this code and examine the details.
+
+.. code-block:: C++
+
+  #include <chrono>
+  #include <functional>
+  #include <string>
+
+  #include <rclcpp/rclcpp.hpp>
+
+  #include <message_filters/delta_filter.hpp>
+  #include <message_filters/subscriber.hpp>
+
+  #include <std_msgs/msg/string.hpp>
+
+  using namespace std::chrono_literals;
+
+  const std::string TUTORIAL_TOPIC_NAME = "tutorial_topic";
+  const rclcpp::QoS QOS_PROFILE = rclcpp::QoS(10);
+
+We start by including ``chrono``, ``string`` and ``functional`` headers.
+The ``chrono`` header is required for the ``chrono_literals`` namespace, necessary for creating timers.
+The ``string`` header provides us with ``std::to_string`` function required for the main functionality demonstration.
+The ``functional`` header is also required to use ``std::bind`` function to bind timer callbacks to timers.
+
+After that we include the ``rclcpp.hpp`` header that provides us with classes from ``rclcpp`` namespace.
+To use filters in our code we need corresponding headers as well.
+In this case we include ``delta_filter.hpp`` for ``DeltaFilter`` and ``subscriber.hpp`` for ``Subscriber`` filter respectively.
+And finally we add ``std_msgs/msg/string.hpp`` to get access to ``String`` message class from the ROS standard messages library.
+
+Lastly, in this section we declare that we are using ``chrono_literals`` namespace.
+We define the ``TUTORIAL_TOPIC_NAME`` and ``QOS_PROFILE`` to use them for subscription and for publishing.
+
+Next we define ``CounterFilter`` class
+
+.. code-block:: C++
+
+  public:
+    typedef message_filters::MessageEvent<std_msgs::msg::String const> EventType;
+
+    CounterFilter() : message_filters::SimpleFilter<std_msgs::msg::String>() {}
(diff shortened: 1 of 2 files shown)


doc/Tutorials/Delta-Filter-Cpp.rst
```diff
@@ -0,0 +1,495 @@
...
+    }
+  private:
+    size_t counter_ = 0;
+
+    message_filters::Connection incoming_connection_;
+
+    void callback(const std_msgs::msg::String::ConstSharedPtr& evt) {
+      add(evt);
+    }
+  };
```
ahcorde, line 81:
```suggestion
  const std::string TUTORIAL_TOPIC_NAME = "tutorial_topic";
  const rclcpp::QoS QOS_PROFILE = rclcpp::QoS(10);

  class CounterFilter : public message_filters::SimpleFilter<std_msgs::msg::String> {
  public:
    typedef message_filters::MessageEvent<std_msgs::msg::String const> EventType;

    CounterFilter() : message_filters::SimpleFilter<std_msgs::msg::String>() {}

    void add(const EventType & evt) {
      ++counter_;
      signalMessage(evt);
    }

    template<class F>
    void connectInput(F & f) {
      incoming_connection_.disconnect();
      incoming_connection_ = f.registerCallback(
        message_filters::SimpleFilter<std_msgs::msg::String>::EventCallback(
          std::bind(
            &CounterFilter::add,
            this,
            std::placeholders::_1
          )
        )
      );
    }

    size_t getCounterValue() {
      return counter_;
    }
  private:
    size_t counter_ = 0;

    message_filters::Connection incoming_connection_;
  };
```

doc/Tutorials/Delta-Filter-Cpp.rst
```diff
@@ -0,0 +1,495 @@
...
+
+    void publisher_timer_callback() {
+      auto message = std_msgs::msg::String();
+      if (pub_count % 2 == 0) {
+        message.data = "pub_count = " + std::to_string(pub_count);
+      } else {
+        message.data = "pub_count = " + std::to_string(pub_count - 1);
+      }
+      publisher_->publish(message);
+      ++pub_count;
```
ahcorde:
```suggestion
      auto message = std_msgs::msg::String();
      if (pub_count_ % 2 == 0) {
        message.data = "pub_count = " + std::to_string(pub_count_);
      } else {
        message.data = "pub_count = " + std::to_string(pub_count_ - 1);
      }
      publisher_->publish(message);
      ++pub_count_;
```

doc/Tutorials/Delta-Filter-Cpp.rst
```diff
@@ -0,0 +1,495 @@
...
+
+    message_filters::DeltaFilter<std_msgs::msg::String> delta_filter_;
+    CounterFilter counter_filter_;
+
+    rclcpp::Publisher<std_msgs::msg::String>::SharedPtr publisher_;
+
+    rclcpp::TimerBase::SharedPtr publisher_timer_;
+    rclcpp::TimerBase::SharedPtr query_timer_;
+
+    int pub_count { 0 };
```
ahcorde:
```suggestion
    int pub_count_ { 0 };
```

doc/Tutorials/Delta-Filter-Cpp.rst
```diff
@@ -0,0 +1,495 @@
...
+
+  #include <message_filters/delta_filter.hpp>
+  #include <message_filters/subscriber.hpp>
+
+  #include <std_msgs/msg/string.hpp>
+
+  using namespace std::chrono_literals;
+
+  const std::string TUTORIAL_TOPIC_NAME = "tutorial_topic";
+  rclcpp::QoS QOS_PROFILE = rclcpp::QoS(10);
```
ahcorde:
```suggestion
  const rclcpp::QoS QOS_PROFILE = rclcpp::QoS(10);
```

doc/Tutorials/Delta-Filter-Cpp.rst
```diff
@@ -0,0 +1,495 @@
...
+
+We start by including ``chrono``, ``string`` and ``functional`` headers.
+The ``chrono`` header is required for the ``chrono_literals`` namespace, necessary for creating timers.
+The ``string`` header provides us with ``std::to_string`` function required for the main functionality demonstration.
+The ``functional`` header is also required to use ``std::bind`` function to bind timer callbacks to timers.
+
+After that we include the ``rclcpp.hpp`` header that provides us with classes from ``rclcpp`` namespace.
+To use filters in our code we need corresponding headers as well.
+In this case we include ``delta_filter.hpp`` for ``DeltaFilter`` and ``subscriber.hpp`` for ``Subscriber`` filter respectively.
+And finally we add ``string.hpp`` to get access to ``String`` message class from the ROS standard messages library.
```
ahcorde:
```suggestion
And finally we add ``std_msgs/msg/string.hpp`` to get access to ``String`` message class from the ROS standard messages library.
```

doc/Tutorials/Delta-Filter-Cpp.rst
```diff
@@ -0,0 +1,495 @@
...
+    }
+  private:
+    size_t counter_ = 0;
+
+    message_filters::Connection incoming_connection_;
+
+    void callback(const std_msgs::msg::String::ConstSharedPtr& evt) {
+      add(evt);
+    }
+  };
```
ahcorde, line 228:
```suggestion
  public:
    typedef message_filters::MessageEvent<std_msgs::msg::String const> EventType;

    CounterFilter() : message_filters::SimpleFilter<std_msgs::msg::String>() {}

    void add(const EventType & evt) {
      ++counter_;
      signalMessage(evt);
    }

    template<class F>
    void connectInput(F & f) {
      incoming_connection_.disconnect();
      incoming_connection_ = f.registerCallback(
        message_filters::SimpleFilter<std_msgs::msg::String>::EventCallback(
          std::bind(
            &CounterFilter::add,
            this,
            std::placeholders::_1
          )
        )
      );
    }

    size_t getCounterValue() {
      return counter_;
    }
  private:
    size_t counter_ = 0;

    message_filters::Connection incoming_connection_;
  };
```

doc/Tutorials/Delta-Filter-Cpp.rst
```diff
@@ -0,0 +1,495 @@
...
+
+This filter counts the number of messages passing through it.
+The ``add`` method increases messages count, and passes message to the following filter via ``signalMessage``.
+The ``connectInput`` connects this filter to a previous filter's output.
+The ``getCounterValue`` grants access to the current messages count.
```
ahcorde:
```suggestion
The ``add`` method increases the message count, and passes the message to the following filter via ``signalMessage``.
The ``connectInput`` connects this filter to a previous filter's output by registering ``add`` as an ``EventCallback`` so the event is forwarded with no extra copies.
The ``getCounterValue`` grants access to the current message count.
```

doc/Tutorials/Delta-Filter-Cpp.rst
```diff
@@ -0,0 +1,495 @@
...
+
+    message_filters::DeltaFilter<std_msgs::msg::String> delta_filter_;
+    CounterFilter counter_filter_;
+
+    rclcpp::Publisher<std_msgs::msg::String>::SharedPtr publisher_;
+
+    rclcpp::TimerBase::SharedPtr publisher_timer_;
+    rclcpp::TimerBase::SharedPtr query_timer_;
+
+    int pub_count { 0 };
```
ahcorde:
```suggestion
    int pub_count_ { 0 };
```

doc/Tutorials/Delta-Filter-Cpp.rst
```diff
@@ -0,0 +1,495 @@
...
+
+We start with a ``Subscriber`` filter, that is going to be an entry point for the messages into our chain of filters.
+After that we add ``DeltaFilter`` and ``CounterFilter`` to the class.
+To publish messages we will need a ``Publisher`` object.
+To automate publishing messages and for querying the chain filter we add two timers, the ``publisher_timer_`` and the ``query_timer_`` respectively.
+
+To demonstrate the functionality of the ``DeltaFilter`` we need to change only some of the messages.
+For that purpose we add ``pub_count``.
```
ahcorde:
```suggestion
For that purpose we add ``pub_count_``.
```

doc/Tutorials/Delta-Filter-Cpp.rst
```diff
@@ -0,0 +1,495 @@
...
+We are going to add the correct value of this counter to every ``even`` message and add the previous value of this counter to every ``odd`` message.
```
ahcorde:
```suggestion
Every even message embeds the current value of this counter, and every odd message re-uses the previous value, so its ``data`` field is identical to its predecessor.
```

doc/Tutorials/Delta-Filter-Cpp.rst
```diff
@@ -0,0 +1,495 @@
...
+    void publisher_timer_callback() {
+      auto message = std_msgs::msg::String();
+      if (pub_count % 2 == 0) {
+        message.data = "pub_count = " + std::to_string(pub_count);
+      } else {
+        message.data = "pub_count = " + std::to_string(pub_count - 1);
+      }
+      publisher_->publish(message);
+      ++pub_count;
+    }
```
ahcorde, line 377:
```suggestion
    void publisher_timer_callback() {
      auto message = std_msgs::msg::String();
      if (pub_count_ % 2 == 0) {
        message.data = "pub_count = " + std::to_string(pub_count_);
      } else {
        message.data = "pub_count = " + std::to_string(pub_count_ - 1);
      }
      publisher_->publish(message);
      ++pub_count_;
    }
```

doc/Tutorials/Delta-Filter-Cpp.rst
```diff
@@ -0,0 +1,495 @@
...
+  [INFO] [1779744361.607395038] [DeltaNode]: Messages passed down the chain count: 3
+
+Note that when the first query is executed, no message have reached the counter filter yet.
+So that is indicated by first line of the console output.
+At the time the next query is executed, the first message has reached the counter.
+So the number of messages is now ``1``.
+Then, when the third query is executed, the third message has been published,
+but was ignored because the data in it was the same as in previous message.
+And so on.
+
```
ahcorde, line 487:
```suggestion

Both timers run at the same 1 s period.
Even though ``publisher_timer_`` is created (and therefore fires) before ``query_timer_``, message delivery from publisher to subscriber is asynchronous, so on the first tick the published message has not yet been routed through the filter chain by the time the query callback runs — which is why the very first line reports ``0``.
From the second tick onward the prior message has been delivered, so the count advances as expected.

```

doc/Tutorials/Delta-Filter-Cpp.rst
```diff
@@ -0,0 +1,495 @@
...
+6. Other options provided by the DeltaFilter package
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Also, if you wish to make your comparison faster and to compare multiple messages with one field_getter call
+you may create your own ``ComparisonHandler``, inheriting from the ``CachedComparisonHandler``
+and define your own comparison rules for it and you should be good to go.
+This ``CustomComparisonHandler`` should implement the ``message_fits`` method that accepts a message as a parameter.
+It should return ``True`` if the message fits your custom condition and ``False`` if it does not.
```
ahcorde:
```suggestion
If you want different comparison semantics than "any field differs", you can plug a custom comparison handler into ``ComparisonFilter``.
The recommended path is to inherit from ``CachedComparisonHandler`` — it already implements the per-field caching and iteration logic in ``message_fits``, and exposes the actual comparison as a pure-virtual hook.

In your subclass, override ``do_fields_fit``:

.. code-block:: C++

  template<typename MessageType>
  class MyCompare : public message_filters::CachedComparisonHandler<MessageType>
  {
  public:
    using FieldGetterFunctionType =
      typename message_filters::CachedComparisonHandler<MessageType>::FieldGetterFunctionType;

    explicit MyCompare(std::forward_list<FieldGetterFunctionType> field_getters)
    : message_filters::CachedComparisonHandler<MessageType>(field_getters) {}

    bool do_fields_fit(message_filters::MFieldType a, message_filters::MFieldType b) const override
    {
      // return true to accept the new message, false to drop it
      return a != b;
    }
  };

``do_fields_fit`` receives the value produced by one of the ``field_getters`` for the cached message (``a``) and for the incoming message (``b``).
Return ``true`` to accept the incoming message (it then becomes the new cached message and is passed down the chain); return ``false`` to drop it.
``DeltaCompare`` itself is just this pattern with ``return a != b;``.

You can then use your handler through ``ComparisonFilter``:

.. code-block:: C++

  message_filters::ComparisonFilter<std_msgs::msg::String, MyCompare> my_filter(
    subscriber_filter_,
    {[] (const MConstPtr & msg) {return msg->data;}}
  );
```

ahcorde (changes requested):

EsipovPA:
@ahcorde  Thank's for the review. I've fixed the places you've pointed to. Could you please have another look when you have the time? Thank you :)

doc/Tutorials/Delta-Filter-Cpp.rst
```diff
@@ -0,0 +1,530 @@
...
+
+Both timers run at the same 1 s period.
+Even though ``publisher_timer_`` is created (and therefore fires) before ``query_timer_``,
+message delivery from publisher to subscriber is asynchronous, so on the first tick the published message
+has not yet been routed through the filter chain by the time the query callback runs — which is why
+the very first line reports ``0``.
+From the second tick onward the prior message has been delivered, so the count advances as expected.
+
+6. Other options provided by the DeltaFilter package
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
ahcorde:
```suggestion
6. Other options provided by the DeltaFilter class                                                                                                                          
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

ahcorde (approved):

ahcorde:
https://github.com/Mergifyio backport lyrical kilted jazzy humble

PR #319, opened 2026-06-23
(#97) MessageTraits: Add support for custom TimeStamp

### LINKED ISSUE #97: Support for custom timestamp
Hello!
Thank you for developing such a useful package. It has become an irreplaceable piece of our system.

In our system, the timestamp is expressed as a single uint64 number with nanosecond precision. Even if converting from `nanosec` to `sec` + `nanosec` is perfectly feasible, it would be cleaner if we could just use the timestamp as is.

Hence my question: would it be possible to use a custom definition of [std_msgs/msg/Header Message](https://docs.ros2.org/foxy/api/std_msgs/index-msg.html) for time synchronization with `TimeSynchronizer` or `ApproximateTimeSynchronizer`?

Concretely, I would like to use a header whose timestamp is a single `uint64` stamp with nanosecond precision, instead of two `int32` and `uint32` stamps with second and nanosecond precision respectively, as in the default ROS2 definition.

If this feature is not implemented yet, I would like to contribute with a pull request. Just let me know if there is any request or suggestion on your side.

Regards
Giorgio

EsipovPA:
## Description

Added support for custom timestamp

Fixes #97 

Yes. Gives user an option to utilize custom timestamp with message_filters.

No.

### DIFF
--- a/include/message_filters/cache.hpp
+++ b/include/message_filters/cache.hpp
@@ -60,7 +60,8 @@ namespace message_filters
 void callback(const std::shared_ptr<M const> &);
 \endverbatim
  */
-template<class M>
+template<class M,
+  template<typename GetterMessageType> typename TimeGetter = message_traits::TimeGetterBase>
 class Cache : public SimpleFilter<M>
 {
 public:
@@ -345,7 +346,8 @@ private:
 
   static rclcpp::Time getMessageTimeFromHeader(const EventType & evt)
   {
-    return message_filters::message_traits::TimeStamp<M>::value(*(evt.getMessage()));
+    return message_filters::message_traits::TimeStampCustom<M,
+             TimeGetter>::value(*(evt.getMessage()));
   }
 
   static rclcpp::Time getMessageReceiveTime(const EventType & evt)
--- a/include/message_filters/cache.hpp
+++ b/include/message_filters/cache.hpp
@@ -61,12 +61,19 @@ void callback(const std::shared_ptr<M const> &);
 \endverbatim
  */
 template<class M,
-  template<typename GetterMessageType> typename TimeGetter = message_traits::TimeGetterBase>
+  template<typename> typename TimeGetter = message_traits::DefaultTimeGetter>
+requires message_traits::TimeGetterFor<TimeGetter, M>
 class Cache : public SimpleFilter<M>
 {
 public:
   using MConstPtr = std::shared_ptr<M const>;
   using EventType = MessageEvent<M const>;
+
+  // A custom TimeGetter extracts the timestamp itself, so the headerless
+  // fallback only applies when the default getter has no header to read.
+  static constexpr bool kMessageProvidesTime =
+    message_traits::HasHeader<M>::value ||
+    !std::is_same_v<TimeGetter<M>, message_traits::DefaultTimeGetter<M>>;
 
   template<class F>
   explicit Cache(F & f, unsigned int cache_size = 1)
@@ -78,7 +85,7 @@ public:
     setCacheSize(cache_size);
     connectInput(f);
 
-    if (message_filters::message_traits::HasHeader<M>::value) {
+    if (kMessageProvidesTime) {
       getMessageTime = getMessageTimeFromHeader;
     } else if (allow_headerless) {
       getMessageTime = getMessageReceiveTime;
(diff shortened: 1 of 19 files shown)


test/time_synchronizer_unittest.cpp
```diff
@@ -35,14 +35,14 @@
...
 #include <rclcpp/rclcpp.hpp>
 
-struct Header
-{
-  rclcpp::Time stamp;
-};
+// struct Header
+// {
+//   rclcpp::Time stamp;
+// };
```
EsipovPA: Forgot to remove this

EsipovPA:
Added support for custom timestamp to Python filters.

EsipovPA:
Considering [this PR](https://github.com/ros2/message_filters/pull/318) and the issue it is solving, I guess I should test this on `RHEL 8 Humble` to check if it is backportable. Will try to do it during the upcoming week. But, nevertheless, I consider it the final version. More or less. So it is ready for a review :)

EsipovPA:
@wjwwood, @ahcorde, @asymingt

Hello guys! Gust a gentle reminder. I would be greatful for a review.

Thank you 🫡

ahcorde (changes requested):
This PR will break downstream packages.

 - Synchronizer<ArppoximateTime<M0, M1>> no longer compiles
```cpp
ApprimateTime(const ApproximateTime & e )
:Parent::ApproximateTIme(e)   // Error
```
 - Copy assigment of all three policy wrappers fails to compile too
 - initParent(void*) 
 - Python: The custom time_getter is never called for the messages that need it

I will create a PR with some ideas.

EsipovPA:
@ahcorde 

Thanks for the review! Seems like I've missed a few things. I'll try to work on a solution.

ahcorde:
@EsipovPA I created some suggestions here https://github.com/EsipovPA/message_filters/pull/1

EsipovPA:
Thank you Alejandro. I've added your changes. Fixing conflicts with `rolling`.

@ahcorde I've managed to resolve conflicts with `rolling`. Please, could you take another look at this PR if you have the time?

include/message_filters/cache.hpp
```diff
@@ -60,13 +60,21 @@ namespace message_filters
...
 {
 public:
   using MConstPtr = std::shared_ptr<M const>;
   using EventType = MessageEvent<M const>;
 
+  // A custom TimeGetter extracts the timestamp itself, so the headerless
+  // fallback only applies when the default getter has no header to read.
+  static constexpr bool kMessageProvidesTime =
+    message_traits::HasHeader<M>::value ||
+    !std::is_same_v<TimeGetter<M>, message_traits::DefaultTimeGetter<M>>;
```
wjwwood, line 76: Is there a problem with this if the user has specialized `message_traits::TimeStamp<UserHeaderlessMsgT>`? Seems to me like the cache might consider it for "`allow_headerless`" despite the specialization, but maybe I'm missing something?

wjwwood (commented):
I had one comment, but otherwise lgtm. I think @ahcorde is out of the office for a bit, but I think you also addressed his concerns (from what I can tell).

include/message_filters/cache.hpp
EsipovPA: Yes, it makes sense to `allow_headerless` for Cache with custom timestamp. I'll try to address it. I need some time to think on a correct approach. At the first glance it seems that the best way to go with that would be to make `allow_headerless` some kind of a `bypass` option to ignore any time getter completely. The syntax will be kinda messy in this case, but it'll be the UNIX-way for this, as I see it. I'll start working on a solution as soon as possible (have some computer time limitations right now)

LastStarDust (approved):
The implementation looks good (very good actually) to me, but I was more concerned about the usability, so I created a small experiment to showcase this feature and it is doing exactly what I need and is very concise and lean.

https://github.com/LastStarDust/ros_experiments/tree/main/src/cpp_time_synchronizer_experiment

PS
I only checked the C++ part.

ahcorde (approved):

ahcorde:
Pulls: ros2/message_filters#319
Gist: https://gist.githubusercontent.com/ahcorde/c59abfc9448ebf907682aa0d8eba99aa/raw/9d46083822d626b13f21f1977ac37a308eade1ce/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/20130
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/30196/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/22894/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/10006/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/29049/)

PR #329, opened 2026-09-03
Fixed doxygen warnings

ahcorde:
## Description

Fixed doxygen warnings

Claude Opus 4.8

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -39,13 +39,29 @@
 # All configuration values have a default; values that are commented out
 # serve to show the default.
 
-# import sys, os
+import os
+import re
+import sys
 from datetime import datetime
 
-# If extensions (or modules to document with autodoc) are in another directory,
-# add these directories to sys.path here. If the directory is relative to the
-# documentation root, use os.path.abspath to make it absolute, like shown here.
-# sys.path.append(os.path.abspath('.'))
+# Put this package's Python sources on sys.path so autodoc can import them.
+#
+# rosdoc2 only does this for packages that let it generate conf.py; this one ships
+# its own, so without this autodoc reports "No module named 'message_filters'"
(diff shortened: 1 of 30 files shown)


ahcorde:
Do you mind to take a look @EsipovPA?

EsipovPA:
@ahcorde Sure thing. Will try to do before the end of the upcoming weekend.

EsipovPA:
Looks fine to me.

But there is one issue. I am not sure if it relates to this specific PR, but I was unable to pinpoint the reason for that myself yet. Maybe this PR is the right place to discuss it? When I build the documentation locally, I get the `Build your package` section in every tutorial as it should be. And in the published version it is somewhat empty. I have no access to logs of the CI that does documentation publishing, so what I am dealing is only the final result. I was wandering, if you could take a look at this specific issue with this PR as well? Or maybe it is better to put up another issue in the issues section? Maybe I'll be able to find a solution to this later. Or does this PR resolve this problem as well?

Here is how it looks when I build it myself
image

Here is what I see in the published version. This is a tutorial for DeltaFileter but others like `Chain` or `Cache` have the same issues. At least when I open the web page.
image

EsipovPA:
@ahcorde  I would approve this PR, I just have no power to do so. :)

ahcorde:
Thank you for the review @EsipovPA

doc/conf.py
```diff
@@ -39,20 +39,36 @@
 # All configuration values have a default; values that are commented out
 # serve to show the default.
 
-# import sys, os
+import os
```
mini-1235, line 42: This seems to introduce the warnings in https://ci.ros2.org/job/ci_linux-aarch64/23222/testReport/junit/(root)/projectroot/flake8/

